PR #3, opened 2022-12-23
Initial Docs

Flydexo:


Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-anyhow = "1.0.65"
-futures = "0.3.25"
-tarpc = { version = "0.30.0", features = ["tokio1"] }
-tokio = { version = "1.21.2", features = ["rt", "rt-multi-thread", "macros"] }
+clap = { version = "4.0.32", features = ["derive"] }
+rust-crypto = "0.2.36"
+jsonrpc-http-server = "18.0"
```
Eikix, line 11: why this framework:)? I understand there are more standard choices for building a json rpc server

Flydexo: `json-http-server` because it is the same lib used by the Parity client, `rust-crypto` for hex, bytes, hashing, etc... and `clap` is required with `json-http-server`

docs/glossary.md
```diff
@@ -0,0 +1,5 @@
+# Glossary
+
+* **Native**: By something native, we mention something living on the starknet network (eg: native ETH = StarknetETH)
+* **KEVM**: By something KEVM, we mention something living on the kakarot contract but with additional admin properties managed by the kakarot contract (eg: KEVM WETH)
+* **EVM**: By something EVM, we mention something living on the kakarot contract (eg: an EVM token)
```
Eikix, line 5: EVM tokens don't live on the Kakarot contract per se:

As of December 2022, they are being deployed on Starknet (through Kakarot) as smart contracts that store their own EVM bytecode. 
They are later on called using an address registry (a separate smart contract holding a 1 to 1 mapping of EVM contracts' EVM addresses <-> starknet addresses). Only Kakarot can interact with this registry contract.

docs/glossary.md
```diff
@@ -0,0 +1,5 @@
+# Glossary
+
+* **Native**: By something native, we mention something living on the starknet network (eg: native ETH = StarknetETH)
+* **KEVM**: By something KEVM, we mention something living on the kakarot contract but with additional admin properties managed by the kakarot contract (eg: KEVM WETH)
```
Eikix, line 4: same here, nothing lives on the Kakarot contract per se, Kakarot only registers the address at which the account registry lives (see below to understand what is the address registry or account registry)

docs/glossary.md
```diff
@@ -0,0 +1,5 @@
+# Glossary
+
+* **Native**: By something native, we mention something living on the starknet network (eg: native ETH = StarknetETH)
```
Eikix, line 3: fine with me:)

Eikix (approved):
lgtm:)

PR #12, opened 2022-12-30
Implementation of a simple Kakarot Server

danilowhk:
Implemented a server that can call "execute" function of Kakarot in Goerli2 Tesnet.

To start the server, run "cargo run" on the root of the repo.

Features Implemented:

- hex_string_to_felt_vec function, which takes a Hex string  as argument and returns <FieldElement>
- call_execute function, which takes bytecode and calldata strings as arguments and call the "execute" in Kakarot.
- Simple rocket server with "kakarot/execute/<tx_bytecode>/<tx_calldata>" endpoint.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,10 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-anyhow = "1.0.65"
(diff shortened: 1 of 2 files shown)


Eikix (commented):
I think this PR implements a REST API

danilowhk:
You are correct, I will change it to RPC and update the PR

PR #34, opened 2023-01-17
feat: Kakarot Rpc method "eth_getCode" + automock test structure

danilowhk:
- eth_blockNumber RPC Method implemented
- Implemented helper function for BlockId/BlockTag conversions
- Added automock test structure

Time spent on this PR:
- blockNumber first implementatino: 1.5 days
- AutoMock test structuring : 0.5 days

# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #20

# What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

# Other information

### DIFF
--- a/core/src/lightclient/mod.rs
+++ b/core/src/lightclient/mod.rs
@@ -1,12 +1,24 @@
 use eyre::Result;
 use jsonrpsee::types::error::CallError;
 use reth_rpc_types::RichBlock;
-use starknet::providers::jsonrpc::models::BlockId as StarknetBlockId;
-use starknet::providers::jsonrpc::{HttpTransport, JsonRpcClient, JsonRpcClientError};
+use starknet::{
+    core::types::FieldElement,
+    macros::selector,
+    providers::jsonrpc::{
+        models::{BlockId as StarknetBlockId, FunctionCall},
+        HttpTransport, JsonRpcClient, JsonRpcClientError,
+    },
+};
 use thiserror::Error;
 use url::Url;
+extern crate hex;
+use reth_primitives::{Address, Bytes};
 
 use crate::helpers::{starknet_block_to_eth_block, MaybePendingStarknetBlock};
+
+use async_trait::async_trait;
+use mockall::predicate::*;
+use mockall::*;
 
 #[derive(Error, Debug)]
 pub enum LightClientError {
@@ -16,8 +28,25 @@ pub enum LightClientError {
     OtherError(#[from] anyhow::Error),
 }
 
-pub struct StarknetClient {
+#[automock]
+#[async_trait]
+pub trait StarknetClient: Send + Sync {
+    async fn block_number(&self) -> Result<u64, LightClientError>;
+    async fn get_eth_block_from_starknet_block(
+        &self,
+        block_id: StarknetBlockId,
+        hydrated_tx: bool,
+    ) -> Result<RichBlock, LightClientError>;
+    async fn get_code(
+        &self,
+        ethereum_address: Address,
+        starknet_block_id: StarknetBlockId,
+    ) -> Result<Bytes, LightClientError>;
+}
+pub struct StarknetClientImpl {
     client: JsonRpcClient<HttpTransport>,
+    // kakarot_contract_address: FieldElement,
+    kakarot_account_registry: FieldElement,
 }
 
 impl From<LightClientError> for jsonrpsee::core::Error {
@@ -31,14 +60,25 @@ impl From<LightClientError> for jsonrpsee::core::Error {
     }
 }
 
-impl StarknetClient {
+impl StarknetClientImpl {
     pub fn new(starknet_rpc: &str) -> Result<Self> {
         let url = Url::parse(starknet_rpc)?;
         Ok(Self {
             client: JsonRpcClient::new(HttpTransport::new(url)),
+            // kakarot_contract_address: FieldElement::from_hex_be(
+            //     "0x031ddf73d0285cc2f08bd4a2c93229f595f2f6e64b25846fc0957a2faa7ef7bb",
+            // )
+            // .unwrap(),
+            kakarot_account_registry: FieldElement::from_hex_be(
+                "0x052a419fd88f53f9a29d22c3d8db24dd9a9a01a41a483ac660d88622f83c40db",
+            )
+            .unwrap(),
         })
     }
+}
 
+#[async_trait]
+impl StarknetClient for StarknetClientImpl {
     /// Get the number of transactions in a block given a block id.
     /// The number of transactions in a block.
     ///
@@ -52,7 +92,7 @@ impl StarknetClient {
     ///
     /// `Ok(ContractClass)` if the operation was successful.
     /// `Err(LightClientError)` if the operation failed.
-    pub async fn block_number(&self) -> Result<u64, LightClientError> {
+    async fn block_number(&self) -> Result<u64, LightClientError> {
         let block_number = self.client.block_number().await?;
         Ok(block_number)
     }
@@ -65,7 +105,7 @@ impl StarknetClient {
     /// # Returns
     /// `Ok(RichBlock)` if the operation was successful.
     /// `Err(LightClientError)` if the operation failed.
-    pub async fn get_eth_block_from_starknet_block(
+    async fn get_eth_block_from_starknet_block(
         &self,
         block_id: StarknetBlockId,
         hydrated_tx: bool,
@@ -81,5 +121,62 @@ impl StarknetClient {
         };
         let block = starknet_block_to_eth_block(starknet_block);
         Ok(block)
+    }
+
+    /// Get the number of transactions in a block given a block id.
+    /// The number of transactions in a block.
+    ///
+    /// # Arguments
+    ///
+    ///
+    ///
+    /// # Returns
+    ///
+    ///  * `block_number(u64)` - The block number.
+    ///
+    /// `Ok(Bytes)` if the operation was successful.
+    /// `Err(LightClientError)` if the operation failed.
+    async fn get_code(
+        &self,
+        ethereum_address: Address,
+        starknet_block_id: StarknetBlockId,
+    ) -> Result<Bytes, LightClientError> {
+        // Convert the Ethereum address to a hex-encoded string
+        let address_hex = hex::encode(ethereum_address);
+        // Convert the hex-encoded string to a FieldElement
+        let ethereum_address_field_element = FieldElement::from_hex_be(&address_hex).unwrap();
+
+        // Prepare the calldata for the get_starknet_contract_address function call
+        let tx_calldata_vec = vec![ethereum_address_field_element];
+        let request = FunctionCall {
+            contract_address: self.kakarot_account_registry,
+            entry_point_selector: selector!("get_starknet_contract_address"),
+            calldata: tx_calldata_vec,
+        };
+        // Make the function call to get the Starknet contract address
+        let starknet_contract_address = self.client.call(request, &starknet_block_id).await?;
+        // Concatenate the result of the function call
+        let concatenated_result = starknet_contract_address
+            .into_iter()
+            .fold(FieldElement::ZERO, |acc, x| acc + x);
+
+        // Prepare the calldata for the bytecode function call
+        let tx_calldata_vec2 = vec![];
+        let request = FunctionCall {
+            contract_address: concatenated_result,
+            entry_point_selector: selector!("bytecode"),
+            calldata: tx_calldata_vec2,
+        };
+        // Make the function call to get the contract bytecode
+        let contract_bytecode = self.client.call(request, &starknet_block_id).await?;
+        // Convert the result of the function call to a vector of bytes
+        let contract_bytecode_in_u8: Vec<u8> = contract_bytecode
+            .into_iter()
+            .flat_map(|x| x.to_bytes_be())
+            .collect();
+        let bytes_result = Bytes::from(contract_bytecode_in_u8);
+
+        // Return the bytecode as a Result<Bytes>
+        Ok(bytes_result)
     }
 }
--- a/core/src/lightclient/mod.rs
+++ b/core/src/lightclient/mod.rs
@@ -20,7 +20,7 @@ use mockall::predicate::*;
 use mockall::*;
 mod constants;
 use constants::{
-    selectors::{SELECTOR_BYTECODE, SELECTOR_GET_STARKNET_CONTRACT_ADDRESS},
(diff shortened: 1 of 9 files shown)


rpc/src/eth_rpc.rs
```diff
@@ -256,6 +256,16 @@ impl EthApiServer for KakarotEthRpc {
         Ok(block_number)
     }
 
+    async fn get_code(&self, _address: Address, _block_number: Option<BlockId>) -> Result<Bytes> {
+        let code = self
+            .starknet_client
+            .get_code(_address, _block_number)
+            .await
+            .map_err(|e| eyre::eyre!(e))
+            .unwrap();
```
Eikix, line 265: Why do we unwrap here? This may panic

rpc/src/eth_rpc.rs
```diff
@@ -256,6 +256,16 @@ impl EthApiServer for KakarotEthRpc {
         Ok(block_number)
     }
 
+    async fn get_code(&self, _address: Address, _block_number: Option<BlockId>) -> Result<Bytes> {
```
Eikix, line 259: What is the format of the Address type? Is it compatible with a Starknet address?

rpc/src/eth_rpc.rs
```diff
@@ -256,6 +256,16 @@ impl EthApiServer for KakarotEthRpc {
...
+        let code = self
+            .starknet_client
+            .get_code(_address, _block_number)
```
Eikix, line 262: This this is calling a Kakarot Contract account, we must consider users input an EVM-format address and then we do a conversion using the kakarot address registry! Then, we use the Starknet address to fetch the code:)

Eikix: Ahhh okay you did it above in the get_code of Starknet client^^

core/src/helpers.rs
```diff
@@ -0,0 +1,27 @@
...
+extern crate hex;
+// use rustc_hex::ToHex;
+
+use reth_primitives::rpc::{BlockId, BlockNumber};
+
+pub fn ethers_block_id_to_starknet_block_id(block: BlockId) -> Result<StarknetBlockId> {
+    match block {
+        BlockId::Hash(hash) => {
+            let address_hex = hex::encode(hash);
+            let address_felt = FieldElement::from_hex_be(&address_hex).unwrap();
```
Eikix, line 15: We should "almost" never use `unwrap`, unless we're certain that the code will not actually error out:)
Generally, if one provides as a comment above the line of code the reason why their code cannot error, i'm ok with unwrap,
otherwise, let's always use error management using `?`

Eikix: I'll create a PR to handle errors I guess!:)

danilowhk: Ok!

core/src/lightclient/mod.rs
```diff
@@ -16,8 +28,25 @@ pub enum LightClientError {
...
+        block_id: StarknetBlockId,
+        hydrated_tx: bool,
+    ) -> Result<RichBlock, LightClientError>;
+    async fn get_code(
+        &self,
+        ethereum_address: Address,
+        starknet_block_id: StarknetBlockId,
+    ) -> Result<Bytes, LightClientError>;
+}
+pub struct StarknetClientImpl {
```
Eikix, line 46: Why does a struct have suffix impl in its name:)?

core/src/lightclient/mod.rs
```diff
@@ -31,14 +60,25 @@ impl From<LightClientError> for jsonrpsee::core::Error {
     }
 }
 
-impl StarknetClient {
+impl StarknetClientImpl {
```
Eikix, line 63: I think it feels a bit bad to write implement an implementation

core/src/lightclient/mod.rs
```diff
@@ -31,14 +60,25 @@ impl From<LightClientError> for jsonrpsee::core::Error {
...
         Ok(Self {
             client: JsonRpcClient::new(HttpTransport::new(url)),
+            // kakarot_contract_address: FieldElement::from_hex_be(
+            //     "0x031ddf73d0285cc2f08bd4a2c93229f595f2f6e64b25846fc0957a2faa7ef7bb",
+            // )
+            // .unwrap(),
+            kakarot_account_registry: FieldElement::from_hex_be(
+                "0x052a419fd88f53f9a29d22c3d8db24dd9a9a01a41a483ac660d88622f83c40db",
+            )
+            .unwrap(),
```
Eikix, line 75: why unwrap?:)

Eikix: i think you can declare it above:
```
let kakarot_account_registry = FieldElement::from_hex_be("0x052a419fd88f53f9a29d22c3d8db24dd9a9a01a41a483ac660d88622f83c40db",)?;
... Ok ... 
```

core/src/lightclient/mod.rs
```diff
@@ -82,4 +122,61 @@ impl StarknetClient {
...
+    /// `Err(LightClientError)` if the operation failed.
+    async fn get_code(
+        &self,
+        ethereum_address: Address,
+        starknet_block_id: StarknetBlockId,
+    ) -> Result<Bytes, LightClientError> {
+        // Convert the Ethereum address to a hex-encoded string
+        let address_hex = hex::encode(ethereum_address);
+        // Convert the hex-encoded string to a FieldElement
+        let ethereum_address_field_element = FieldElement::from_hex_be(&address_hex).unwrap();
```
Eikix, line 147: why unwrap:)

Eikix, line 147: if you'd like we can rename everything that is a field element conversion with a _fe suffix
like `ethereum_address_fe`, `block_number_fe` etc. that way we have a unified way to name things

core/src/lightclient/mod.rs
```diff
@@ -82,4 +122,61 @@ impl StarknetClient {
...
+
+        // Prepare the calldata for the get_starknet_contract_address function call
+        let tx_calldata_vec = vec![ethereum_address_field_element];
+        let request = FunctionCall {
+            contract_address: self.kakarot_account_registry,
+            entry_point_selector: selector!("get_starknet_contract_address"),
```
Eikix, line 153: let's create a `selectors.rs` in this folder with only `pub const` declarations of selectors that we may need (we'll add them on one by one as we go) so that we're "type-safe" and centralise all selectors
wdyt

core/src/lightclient/mod.rs
```diff
@@ -82,4 +122,61 @@ impl StarknetClient {
...
+            calldata: tx_calldata_vec,
+        };
+        // Make the function call to get the Starknet contract address
+        let starknet_contract_address = self.client.call(request, &starknet_block_id).await?;
+        // Concatenate the result of the function call
+        let concatenated_result = starknet_contract_address
```
Eikix, line 159: if this is going to be reused at every .call of the client, let's make this into a named util:)! 
I understand that the code of the contract is given in an array of 16bytes chunks that we need to concatenate, right?
let's make a pure function util and test it

core/src/lightclient/mod.rs
```diff
@@ -82,4 +122,61 @@ impl StarknetClient {
...
+            .into_iter()
+            .fold(FieldElement::ZERO, |acc, x| acc + x);
+
+        // Prepare the calldata for the bytecode function call
+        let tx_calldata_vec2 = vec![];
+        let request = FunctionCall {
+            contract_address: concatenated_result,
+            entry_point_selector: selector!("bytecode"),
```
Eikix, line 167: selector inside selectors.rs:)!

later one we could even create script that takes an ABI and generates starknet-rs selectors:)! 
but later!

core/src/lightclient/mod.rs
```diff
@@ -82,4 +122,61 @@ impl StarknetClient {
...
+            calldata: tx_calldata_vec2,
```
Eikix, line 168: if this is empty by design, we could just directly write `vec![]`

core/src/lightclient/mod.rs
```diff
@@ -82,4 +122,61 @@ impl StarknetClient {
...
+        };
+        // Make the function call to get the contract bytecode
+        let contract_bytecode = self.client.call(request, &starknet_block_id).await?;
+        // Convert the result of the function call to a vector of bytes
+        let contract_bytecode_in_u8: Vec<u8> = contract_bytecode
```
Eikix, line 173: if this is going to be reused at every .call of the client, let's make this into a named util:)! 
I understand that the code of the contract is given in an array of 16bytes chunks that we need to concatenate, right?
let's make a pure function util and test it

core/src/lightclient/mod.rs
Eikix, line 159: let's make it into a unique util and test it;) if it'll be reused in every .call of starknetclient! 

unless starknet-rs already has a util.

later on, we can even make a PR on the repo if this util does not exist yet

core/tests/lightclient.rs
```diff
@@ -0,0 +1,80 @@
...
+        // When
+        let result = starknet_lightclient_mock
+            .get_code(ethereum_address, starknet_block_id)
+            .await;
+        assert!(result.is_ok());
+
+        let result_value = result.unwrap();
+        // Then
+        assert_eq!(
+            format!("{bytes_result_value:?}"),
```
Eikix, line 60: why do we need format here:)?

rpc/src/main.rs
```diff
@@ -18,9 +18,9 @@ async fn main() -> Result<()> {
     let starknet_rpc = std::env::var("STARKNET_RPC_URL")
         .map_err(|_| eyre!("Missing mandatory environment variable: STARKNET_RPC_URL"))?;
 
-    let starknet_lightclient = StarknetClient::new(&starknet_rpc)?;
+    let starknet_lightclient = StarknetClientImpl::new(&starknet_rpc)?;
 
-    let (server_addr, server_handle) = run_server(starknet_lightclient).await?;
+    let (server_addr, server_handle) = run_server(Box::new(starknet_lightclient)).await?;
```
Eikix, line 23: why do we need to box it?

Eikix (changes requested):
awesome PR! this sets a cool example for tests on the kakarot-rpc repo!

Eikix:
closes #40

core/src/lightclient/mod.rs
```diff
@@ -31,14 +60,25 @@ impl From<LightClientError> for jsonrpsee::core::Error {
...
     pub fn new(starknet_rpc: &str) -> Result<Self> {
         let url = Url::parse(starknet_rpc)?;
         Ok(Self {
             client: JsonRpcClient::new(HttpTransport::new(url)),
+            // kakarot_contract_address: FieldElement::from_hex_be(
+            //     "0x031ddf73d0285cc2f08bd4a2c93229f595f2f6e64b25846fc0957a2faa7ef7bb",
+            // )
+            // .unwrap(),
+            kakarot_account_registry: FieldElement::from_hex_be(
+                "0x052a419fd88f53f9a29d22c3d8db24dd9a9a01a41a483ac660d88622f83c40db",
```
Eikix, line 73: let's also put this long value as a top-level const:)

Eikix:
```rust
impl StarknetClient {
    pub fn new(starknet_rpc: &str) -> Result<Self> {
        let url = Url::parse(starknet_rpc)?;
        let kakarot_account_registry = FieldElement::from_hex_be(
            ACCOUNT_REGISTRY_ADDRESS,
        )?;
        Ok(Self {
            client: JsonRpcClient::new(HttpTransport::new(url)),
            kakarot_account_registry,
        })
    }
```

core/src/lightclient/mod.rs
danilowhk: Its just to separate from the trait, it is a pattern that was implemented in Beerus to

core/src/lightclient/mod.rs
danilowhk: Hahaha, we can change, but I dont see an issue

core/src/lightclient/mod.rs
danilowhk: Yes, we can use the "?"

core/src/lightclient/mod.rs
danilowhk: Same

danilowhk: Ok

rpc/src/main.rs
danilowhk: Because of automock implementation. Bot exactly sure, but it requires it to be passed as a dyn and a good way was to use Box

core/src/lightclient/mod.rs
danilowhk: Agree, this will def make it a lot safer!

core/src/lightclient/mod.rs
danilowhk: Agree hehe

core/src/lightclient/mod.rs
danilowhk: I didn't get what you mean in this comment. What to u refer to when u mean "going to be reused" ?

core/tests/lightclient.rs
```diff
@@ -0,0 +1,80 @@
+mod test {
```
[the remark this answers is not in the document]
danilowhk: U mean the difference between "rpc" and "rpc_core" tests? The idea for testing in "rpc" is to test the wrapper

core/src/lightclient/mod.rs
danilowhk: Ok!

core/src/lightclient/mod.rs
Eikix: alright let's keep it for now;D

core/src/lightclient/mod.rs
Eikix: if we're not doing it here, we can create an issue for it I guess

core/src/lightclient/mod.rs
Eikix: Nevermind:)! Let's leave it like that, next time, if we have to rewrite the same code (duplication), we can refactor it then

core/src/lightclient/mod.rs
Eikix: Nevermind:)! Let's leave it like that, next time, if we have to rewrite the same code (duplication), we can refactor it then

rpc/src/main.rs
Eikix: Ok!

Eikix (approved):
lgtm

Issue #40, opened 2023-01-20
feat: add testing management to the repo


Eikix:
# Feature Request

**Describe the Feature Request**

**Describe Preferred Solution**

**Describe Alternatives**

**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

PR #42, opened 2023-01-24
Implement block_with_full_txs

danilowhk:
- Implement get_block_with_full_tx
- Add test for get_block_with_tx_hashes

Time spent on this PR: 1

# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #40 

# What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

# Other information

### DIFF
--- a/core/src/helpers.rs
+++ b/core/src/helpers.rs
@@ -1,10 +1,9 @@
-use std::collections::BTreeMap;
-
-use eyre::Result;
+use eyre::{ErrReport, Error, Result};
 use reth_primitives::{
     rpc::{BlockId as EthBlockId, BlockNumber},
     Bloom, Bytes, H160, H256, H64, U256,
 };
+use std::collections::BTreeMap;
 
 use reth_primitives::Address;
 use reth_rpc_types::{
@@ -20,6 +19,17 @@ use starknet::{
 
 use crate::lightclient::LightClientError;
 extern crate hex;
+
+#[derive(Debug)]
+struct InvalidFieldElementError;
+
+impl std::error::Error for InvalidFieldElementError {}
+
+impl std::fmt::Display for InvalidFieldElementError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "Invalid FieldElement")
+    }
+}
 
 pub enum MaybePendingStarknetBlock {
     BlockWithTxHashes(MaybePendingBlockWithTxHashes),
@@ -184,6 +194,7 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
                         .transactions
                         .into_iter()
                         .map(starknet_tx_into_eth_tx)
+                        .filter_map(Result::ok)
                         .collect(),
                 );
                 let header = Header {
@@ -236,6 +247,7 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
                         .transactions
                         .into_iter()
                         .map(starknet_tx_into_eth_tx)
+                        .filter_map(Result::ok)
                         .collect(),
                 );
                 let header = Header {
@@ -276,7 +288,7 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
     }
 }
 
-fn starknet_tx_into_eth_tx(tx: StarknetTransaction) -> EtherTransaction {
+fn starknet_tx_into_eth_tx(tx: StarknetTransaction) -> Result<EtherTransaction, Error> {
     let mut ether_tx = EtherTransaction::default();
 
     match tx {
@@ -290,9 +302,9 @@ fn starknet_tx_into_eth_tx(tx: StarknetTransaction) -> EtherTransaction {
                     // Define gas_price data
                     ether_tx.gas_price = None;
                     // Extracting the signature
-                    ether_tx.r = field_option_element_to_u256(v0.signature.get(0));
-                    ether_tx.s = field_option_element_to_u256(v0.signature.get(1));
-                    ether_tx.v = field_option_element_to_u256(v0.signature.get(2));
+                    ether_tx.r = field_option_element_to_u256(v0.signature.get(0))?;
+                    ether_tx.s = field_option_element_to_u256(v0.signature.get(1))?;
+                    ether_tx.v = field_option_element_to_u256(v0.signature.get(2))?;
                     // Extracting the data (transform from calldata)
                     ether_tx.input = vec_felt_to_bytes(v0.calldata);
                     //TODO:  Fetch transaction To
@@ -320,9 +332,9 @@ fn starknet_tx_into_eth_tx(tx: StarknetTransaction) -> EtherTransaction {
                     // Define gas_price data
                     ether_tx.gas_price = None;
                     // Extracting the signature
-                    ether_tx.r = field_option_element_to_u256(v1.signature.get(0));
-                    ether_tx.s = field_option_element_to_u256(v1.signature.get(1));
-                    ether_tx.v = field_option_element_to_u256(v1.signature.get(2));
+                    ether_tx.r = field_option_element_to_u256(v1.signature.get(0))?;
+                    ether_tx.s = field_option_element_to_u256(v1.signature.get(1))?;
+                    ether_tx.v = field_option_element_to_u256(v1.signature.get(2))?;
                     // Extracting the data
                     ether_tx.input = vec_felt_to_bytes(v1.calldata);
                     // Extracting the to address
@@ -378,9 +390,9 @@ fn starknet_tx_into_eth_tx(tx: StarknetTransaction) -> EtherTransaction {
             // Define gas_price data
             ether_tx.gas_price = None;
             // Extracting the signature
-            ether_tx.r = field_option_element_to_u256(declare_tx.signature.get(0));
-            ether_tx.s = field_option_element_to_u256(declare_tx.signature.get(1));
-            ether_tx.v = field_option_element_to_u256(declare_tx.signature.get(2));
+            ether_tx.r = field_option_element_to_u256(declare_tx.signature.get(0))?;
+            ether_tx.s = field_option_element_to_u256(declare_tx.signature.get(1))?;
+            ether_tx.v = field_option_element_to_u256(declare_tx.signature.get(2))?;
             // Extracting the to address
             ether_tx.to = None;
             // Extracting the value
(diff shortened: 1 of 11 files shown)


core/src/helpers.rs
```diff
@@ -169,16 +172,281 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+            match invoke_tx {
+                InvokeTransaction::V0(v0) => {
+                    // Extract relevant fields from InvokeTransactionV0 and convert them to the corresponding fields in EtherTransaction
+                    ether_tx.hash = H256::from_slice(&v0.transaction_hash.to_bytes_be());
+                    ether_tx.nonce = field_element_to_u256(v0.nonce);
+                    ether_tx.from = starknet_address_to_ethereum_address(v0.contract_address);
+                    // Define gas_price data
+                    ether_tx.gas_price = None;
+                    // Extracting the signature
+                    ether_tx.r = field_element_to_u256(v0.signature[0]);
```
Eikix, line 293: to avoid possible panic, let's use `v0.signature.get(0)`, it'll return an option, from there, we can choose how to handle the `None` case properly

Eikix: wdyt, we can choose to return an custom error, a default 0 value, etc.

core/src/helpers.rs
```diff
@@ -169,16 +172,281 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+                    ether_tx.s = field_element_to_u256(v0.signature[1]);
+                    ether_tx.v = field_element_to_u256(v0.signature[2]);
+                    // Extracting the data (transform from calldata)
+                    ether_tx.input = vec_felt_to_bytes(v0.calldata);
+                    // How to fetch to?
```
Eikix, line 298: Let's replace questions by `TODO: `

core/src/helpers.rs
```diff
@@ -169,16 +172,281 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+                    ether_tx.to = None;
+                    // How to fetch the value?
+                    ether_tx.value = U256::from(0);
+                    // How to fetch gas?
+                    ether_tx.gas = U256::from(0);
+                    // Extracting the chain_id
+                    ether_tx.chain_id = None;
```
Eikix, line 305: chain_id can be a constant. in fact it can be a workspace-level constant that is used in eth_chainId and here! you can perform a quick refactor if you'd like

core/src/helpers.rs
```diff
@@ -169,16 +172,281 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+fn vec_felt_to_bytes(contract_bytecode: Vec<FieldElement>) -> Bytes {
+    let contract_bytecode_in_u8: Vec<u8> = contract_bytecode
+        .into_iter()
+        .flat_map(|x| x.to_bytes_be())
+        .collect();
+    Bytes::from(contract_bytecode_in_u8)
+}
+
+fn starknet_address_to_ethereum_address(x: FieldElement) -> Address {
+    let inner: u64 = x.to_string().parse().unwrap();
```
Eikix, line 446: let's not unwrap:)!

also, it's probably safer to go from felt to bytes then Address from Bytes:)!

Eikix (changes requested):
awesome work!

core/tests/lightclient.rs
```diff
@@ -60,7 +64,82 @@ mod test {
...
+            size,
+        };
+        let expected_result = Rich::<Block> {
+            inner: block,
+            extra_info: BTreeMap::default(),
+        };
+
+        let expected_result_value = expected_result.clone();
+
+        starknet_lightclient_mock
```
Eikix, line 124: isn't this test testing that a clone of `expected_result` is equal to another clone of `expected_result`?

Eikix: I think what this test might need to perform is:
Construct a "dummy" StarknetBlock with txHashes and with tx
then pass them inside `from_starknet_transaction_ethers_transaction`

and assert that the returned transaction is an ethereum tx

core/src/helpers.rs
```diff
@@ -169,16 +172,281 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+                Rich::<Block> {
+                    inner: block,
+                    extra_info: BTreeMap::default(),
+                }
             }
         },
     }
 }
 
+fn from_starknet_transaction_ethers_transaction(tx: StarknetTransaction) -> EtherTransaction {
```
Eikix, line 279: how about `from_starknet_tx_to_eth_tx` 
or `starknet_tx_into_eth_tx` 
i.e. we don't need "ethers" with an s
we don't need to fully spell transaction imho

core/src/helpers.rs
danilowhk: Yeap, makes sense

core/src/helpers.rs
danilowhk: Ok!

core/src/helpers.rs
danilowhk: Ok!

core/src/helpers.rs
danilowhk: Didnt get the second comment, this function is to get from Felt to Address, so got a bit confused

core/tests/lightclient.rs
danilowhk: Oh no, I am just using expect_result_value, because expected_result ownership moves when we do the mocking, so its just a clone to compare the end result with "expected_result" 

I believe for the second point would be to test "from_starknet_transaction_ethers_transaction", but I think I this should be another test, so I will add one there.

core/src/helpers.rs
danilowhk: Much better haha

danilowhk: I like starknet_tx_into_eth_tx 
(Im bad at naming btw)

core/src/helpers.rs
Eikix: Something like `let inner: H160 = H160::from_slice(x.to_bytes_be());` has lesser chances to panic, as u64 is 64 bits integer, if x > u64::MAX(), code will panic

core/tests/lightclient.rs
Eikix: Yes makes sense, I'm just saying `.expect_get_eth_block_from_starknet_block()
            .returning(move |_, _| Ok(expected_result.clone()));` makes it so that the `.get_eth_block_from_starknet_block` method does nothing except return expected_result.clone(), which means that we're not testing anything here;)

danilowhk: Yes, this is how automock works. In case we wanted to test the real implementation, we would need an RPC Link and the tests would be a bit different

core/src/helpers.rs
```diff
@@ -169,16 +172,290 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+fn starknet_tx_into_eth_tx(tx: StarknetTransaction) -> EtherTransaction {
+    let mut ether_tx = EtherTransaction::default();
+
+    match tx {
+        StarknetTransaction::Invoke(invoke_tx) => {
+            match invoke_tx {
+                InvokeTransaction::V0(v0) => {
+                    // Extract relevant fields from InvokeTransactionV0 and convert them to the corresponding fields in EtherTransaction
+                    ether_tx.hash = H256::from_slice(&v0.transaction_hash.to_bytes_be());
+                    ether_tx.nonce = field_element_to_u256(v0.nonce);
```
Eikix, line 288: how about `felt_to_u256`
and `felt_option_to_u256`

core/src/helpers.rs
```diff
@@ -169,16 +172,290 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+                    // Extracting the data (transform from calldata)
+                    ether_tx.input = vec_felt_to_bytes(v0.calldata);
+                    //TODO:  Fetch transaction To
+                    ether_tx.to = None;
+                    //TODO:  Fetch value
+                    ether_tx.value = U256::from(0);
+                    //TODO: Fetch Gas
+                    ether_tx.gas = U256::from(0);
+                    // Extracting the chain_id
+                    ether_tx.chain_id = Some(1263227476_u64.into());
```
Eikix, line 305: We could put this constant in constants.rs, what do you think?

core/src/helpers.rs
```diff
@@ -169,16 +172,290 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+        }
+    }
+
+    ether_tx
+}
+
+fn field_option_element_to_u256(element: Option<&FieldElement>) -> U256 {
+    match element {
+        Some(x) => {
+            let inner: u64 = x.to_string().parse().unwrap();
```
Eikix, line 435: this might panic:) 
no `unwrap` and no `u64` intermediary format 

In rust, u64 is 64 bits, whereas felt is potentially 251 bits
this means we should never use u64 as intermediary format but only bytes

core/src/helpers.rs
```diff
@@ -169,16 +172,290 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
...
+            U256::from(inner)
+        }
+        None => U256::from(0),
```
Eikix, line 438: are there cases where this conversion to "0" value is problematic?
for instance if a signature is missing in `s = deploy_account_tx.signature.get(1)`, we just convert it to 0? we might want to error out

core/tests/lightclient.rs
Eikix: My point is that this test is not testing anything, we can delete it and instead test helpers functions:)!

Then, we only have to make sure that the starknet_client `get_eth_block_from_starknet_block()` has no logic beyond assembling/rearranging our helpers functions

Eikix (changes requested):
A couple of comments and good to go:)!

core/src/helpers.rs
danilowhk: Changed to returning Error

core/src/helpers.rs
danilowhk: Didnt know that! Changed to transforming to bytes and then to U256 :)

core/src/helpers.rs
danilowhk: Ok!

core/src/helpers.rs
danilowhk: Yes, makes sense. (Implemented)

core/src/helpers.rs
danilowhk: Changed :)

core/src/helpers.rs
danilowhk: Done

core/src/helpers.rs
danilowhk: Changed to Return Error

core/tests/lightclient.rs
danilowhk: Ok! May I implement this change on the next PR? I will also add the tests for the helper functions

Eikix: Awesome!

Eikix (approved):
lgtm:)!

PR #60, opened 2023-02-03
feature: max_priority_fee_per_gas

MartianGreed:
# max_priority_fee_per_gas RPC method call

Time spent on this PR : 2 hours

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

### DIFF
--- a/core/src/client/mod.rs
+++ b/core/src/client/mod.rs
@@ -10,7 +10,7 @@ use starknet::{
     core::types::FieldElement,
     providers::jsonrpc::{
         models::{
-            BlockId as StarknetBlockId, BroadcastedInvokeTransaction,
+            BlockId as StarknetBlockId, BlockTag, BroadcastedInvokeTransaction,
             BroadcastedInvokeTransactionV1, DeployAccountTransactionReceipt,
             DeployTransactionReceipt, FunctionCall, InvokeTransaction, InvokeTransactionReceipt,
             MaybePendingBlockWithTxHashes, MaybePendingTransactionReceipt, SyncStatusType,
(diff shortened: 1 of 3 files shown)


MartianGreed:
We need to make an improvement on the fee calculation.

For the moment the computation is the following :
- fetch latest block
- grab 10 transactions (can be parameterized)
- SUM(max_fee) / transaction count.

We discussed with @Eikix about 2 things. We can either inspire on the [EIP 1559](https://www.blocknative.com/blog/eip-1559-fees) or just return like a big number.

Eikix (changes requested):
We might need to rethink how we want to indicate the "max_priority_fee_per_gas" as Starknet currently does not have EIP1559 enabled nor a similar gas market,
It's unclear what the best course of action is. One possibility is to return a fixed amount, very high.


Here is my view on this
In Eth context, there is gas_fee := baseFee + priorityFee

Kakarot rpc can either 
-> in its full blocks rpc methods (getBlockByHash), return a baseFee = to the robust mean of the average of starknet fees for a block and then max_priority_fee could be +10% or +20% on this baseFee amount (to be sure to be included)
This is hard, as we would need to compute a statistical gas used for kakarot transaction thanks to simulation and averages, and then divide max_fee / gas_used -> baseFee

-> OR, can return baseFee == 0 and write max_priority_fee = to the robust mean of the average of starknet fees for a block 

I think the first option is a bit better

Eikix:
I'm thinking about this PR and wondering about the way to go!

MartianGreed:
did @phklive some work on it ? Maybe chat about it on telegram ?

phklive:
> did @phklive work on it ? Maybe chat about it on telegram ?

I have been focused on updating the beerus api actively lately, i haven't been able to make progress.

If you or someone wants to take the hand on this issue, please do so.

Eikix:
I think we should close this PR for a little while and come back to it when we have a sound approach on gas in the Kakarot RPC node.

Let's discuss it during the community call tomorrow?

phklive:
> I think we should close this PR for a little while and come back to it when we have a sound approach on gas in the Kakarot RPC node.
> 
> 
> 
> Let's discuss it during the community call tomorrow?

agreed

Eikix:
Closing this PR optimistically, waiting for more elements to reopen and debate with you:)

Issue #18, opened 2023-01-11
feat: Kakarot Rpc method "eth_getTransactionReceipt"


danilowhk:
# eth_getTransactionReceipt

## Metadata

- name: getTransactionReceipt
- prefix: eth
- state: ⚠️
- [specification](https://github.com/ethereum/execution-apis/blob/6709c2a795b707202e93c4f2867fa0bf2640a84f/src/eth/transaction.yaml#L42)

## Specification Description

Describe the method

### Parameters

- Address : address
- Block : BlockTag

### Returns

- BlockCount : uint

## Kakarot Logic

Describe the interaction with Kakarot and/or Starknet.
e.g. "this method does not interact with the kakarot contract or any other starknet contract."

### Kakarot methods

Which Kakarot methods are called?

### Starknet methods

Which Starknet RPC methods are called? e.g. "Would be calling the corresponding `starknet_getTransactionReceipt` https://github.com/starkware-libs/starknet-specs/blob/master/starknet_vs_ethereum_node_apis.md"

Eikix:
Is someone on this issue?

irisdv:
Yes

PR #67, opened 2023-02-04
Feat: implement eth_getTransactionReceipt

irisdv:
Implement eth_getTransactionReceipt

Time spent on this PR: 6h

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Unimplemented

Issue Number: #18 

# What is the new behavior?

Returns the receipt of a transaction by transaction hash.

Currently it returns : 
- transactionHash
- blockHash
- blockNumber
- from
- to
- contractAddress
- status

And fixed data or default values for : 
- cumulativeGasUsed
- logs
- logsBloom
- root
- effectiveGasPrice
- gasUsed

- [ ] Yes
- [x] No


core/src/lightclient/mod.rs
```diff
@@ -354,4 +367,130 @@ impl StarknetClient for StarknetClientImpl {
...
+                            actual_fee,
+                            status,
+                            block_hash,
+                            block_number,
+                            ..
+                        },
+                    ) => {
+                        res_receipt.transaction_hash =
+                            Some(PrimitiveH256::from_slice(&transaction_hash.to_bytes_be()));
+                        res_receipt.gas_used = Some(felt_to_u256(actual_fee));
```
Eikix, line 446: `actual_fee` is not exactly `gas_used` 
https://docs.starknet.io/documentation/architecture_and_concepts/Fees/fee-mechanism/

I think actual_fee is gas_used * fee

core/src/lightclient/mod.rs
```diff
@@ -354,4 +367,130 @@ impl StarknetClient for StarknetClientImpl {
...
+                        res_receipt.status_code = match status {
+                            StarknetTransactionStatus::Pending => Some(U64::from(0)),
+                            StarknetTransactionStatus::AcceptedOnL1 => Some(U64::from(1)),
+                            StarknetTransactionStatus::AcceptedOnL2 => Some(U64::from(1)),
+                            StarknetTransactionStatus::Rejected => Some(U64::from(0)),
+                        };
+                        res_receipt.block_hash =
+                            Some(PrimitiveH256::from_slice(&block_hash.to_bytes_be()));
```
Eikix, line 454: Nice workaround with PrimitiveH256, could you create an issue to prepare the time where we'll unify our type handling? 

What I mean is an issue that looks like: Go from PrimitiveH256 and H256 => only one H256 type (when reth is updated probably)

core/src/lightclient/mod.rs
```diff
@@ -354,4 +367,130 @@ impl StarknetClient for StarknetClientImpl {
...
+            StarknetTransaction::Invoke(invoke_tx) => {
+                match invoke_tx {
+                    InvokeTransaction::V0(v0) => {
+                        res_receipt.contract_address =
+                            Some(starknet_address_to_ethereum_address(v0.contract_address));
+                    }
+                    InvokeTransaction::V1(_) => {}
+                };
+            }
+            StarknetTransaction::L1Handler(l1_handler_tx) => {
```
Eikix, line 475: It's not clear to me whether or not we can interact with L1 - L2 messaging.

I would suggest adding a TODO comment above that refers to Kakarot not handling L1-L2 messaging

core/src/lightclient/mod.rs
```diff
@@ -354,4 +367,130 @@ impl StarknetClient for StarknetClientImpl {
...
+            }
+        };
+
+        let starknet_tx = self.client.get_transaction_by_hash(hash_felt).await?;
+        match starknet_tx.clone() {
+            StarknetTransaction::Invoke(invoke_tx) => {
+                match invoke_tx {
+                    InvokeTransaction::V0(v0) => {
+                        res_receipt.contract_address =
+                            Some(starknet_address_to_ethereum_address(v0.contract_address));
```
Eikix, line 470: My understanding is that we want to get the EVM address of a Kakarot contract (be it an ExternallyOwnedAccount or a Contract account). At first we simply took the last 20 bytes to represent it as an EVM address. But actually now that I think about it, all Kakarot contracts (EVM smart contracts or accounts) have a view method called `get_evm_address()`

So we should replace occurrences of `starknet_address_to_ethereum_address`(which is slicing) to another function. Let's create an issue for this for now. I'll think about it further and detail more strategy in an issue

core/src/lightclient/types.rs
```diff
@@ -184,3 +184,34 @@ impl<T: Serialize> Serialize for Rich<T> {
         }
     }
 }
+
+// TransactionReceipt
+pub struct MyTransactionReceipt(pub TransactionReceipt);
```
Eikix, line 189: I would suggest against naming this struct `MyTransactionReceipt` => might prefer
`KakarotTransactionReceipt` or `TransactionReceiptWrapper`

Eikix: That being said, it's probably simpler to create a util function, name `create_default_transaction_receipt()` that returns a `TransactionReceipt`

There is no need to create a struct here:)

Eikix (changes requested):
Nice PR! Very important RPC method :))

core/src/lightclient/mod.rs
```diff
@@ -354,4 +367,131 @@ impl StarknetClient for StarknetClientImpl {
...
+                match receipt {
+                    StarknetTransactionReceipt::Invoke(InvokeTransactionReceipt {
+                        transaction_hash,
+                        actual_fee,
+                        status,
+                        block_hash,
+                        block_number,
+                        ..
+                    })
+                    | StarknetTransactionReceipt::L1Handler(L1HandlerTransactionReceipt {
```
Eikix, line 410: I think you can refactor this `match` to work without `|` but rather with different matching arms.

We should throw an error for now on L1Handler, DeclareTransaction, as we don't support them

core/src/lightclient/mod.rs
Eikix: I stand corrected, it's not that we should replace occurrences of `starnet_address_to_ethereum_address`, we should refactor this function to execute the logic we want it to. Could you create an issue for this specific refactor?

core/src/lightclient/mod.rs
Eikix: Good for me!

Eikix (approved):
lgtm! could you create an issue to write down Kakarot "logs" into the transaction_receipt.logs? i think it's a whole Pr in itself as it's a non-trivial task to convert starknet transaction receipt logs to be EVM-format 'friendly'

Eikix:
Conflicts and good to go

Eikix (approved):
lgtm!

Eikix (approved):
lgtm

PR #69, opened 2023-02-04
Added block hash and block number from transaction receipt when converting starknet to ethereum transaction

alex-sumner:
Time spent on this PR: 4 hours

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

No block hash or block number information were included when converting starknet to ethereum transaction and hence were also not included in information returned by eth_getTransactionBy... functions

Issue Number: N/A

# What is the new behavior?

Block hash and number are now returned from transaction receipt (in case of non-pending block, for a pending block they are not available)
-
-
-

- [ ] Yes
- [x] No

# Other information


core/src/helpers.rs
```diff
@@ -449,7 +455,11 @@ pub fn decode_execute_at_address_return(
 
 pub fn starknet_tx_into_eth_tx(
     tx: StarknetTransaction,
+    block_hash: FieldElement,
+    block_number: u64,
```
Eikix, line 459: let's make this a `U256` type

core/src/helpers.rs
```diff
@@ -449,7 +455,11 @@ pub fn decode_execute_at_address_return(
 
 pub fn starknet_tx_into_eth_tx(
     tx: StarknetTransaction,
+    block_hash: FieldElement,
```
Eikix, line 458: let's make both `block_hash` and `block_number` Option types, that way they can be set to None instead of Zeroes when they're not available

core/src/helpers.rs
```diff
@@ -226,7 +226,7 @@ pub fn starknet_block_to_eth_block(block: MaybePendingStarknetBlock) -> RichBloc
                     pending_block_with_txs
                         .transactions
                         .into_iter()
-                        .map(starknet_tx_into_eth_tx)
+                        .map(|t| starknet_tx_into_eth_tx(t, FieldElement::ZERO, 0))
```
Eikix, line 229: Let's change this to `None` in case it's not available

Eikix (changes requested):
Hey:)! Thanks for the PR, and great idea.
After comments are incorporated, ready to merge

core/src/helpers.rs
alex-sumner: Done.

core/src/helpers.rs
alex-sumner: Done

core/src/helpers.rs
alex-sumner: Done

Eikix:
You'll need to rebase on latest main to resolve conflicts

PR #85, opened 2023-02-10
Feat: implement kakarot_getTokenBalances

### LINKED ISSUE #46: feat: kakarot_getTokenBalances
# kakarot_getTokenBalances

## Metadata

- name: getTokenBalances
- prefix: kakarot
- state: ⚠️
- [specification](https://github.com/alchemyplatform/alchemy-web3#web3alchemygettokenbalancesaddress-contractaddresses)

## Specification Description

Returns token balances for a specific address given a list of contracts. 

### Parameters

- `address`: The address for which token balances will be checked.
- `contractAddresses`: An array of contract addresses. 

### Returns

An object with the following fields:

- `address`: The address for which token balances were checked.
- `tokenBalances`: An array of token balance objects. Each object contains:
    - `contractAddress`: The address of the contract.
    - `tokenBalance`: The balance of the contract, as a string representing a base-10 number.
    - `error`: An error string. Either `error` or `tokenBalance` will be null.

## Kakarot Logic

Calls Kakarot execute_at_address with "call signature" once for each `contractAddresses` index -> calls `balanceOf(address owner)` with parameter `owner == address`.
Reconstructs the return format.

### Kakarot methods

Call `execute_at_address` n times,`n == contractAddresses.len()`

irisdv:
Implement kakarot_getTokenBalances, which returns token balances for a specific address given a list of contracts.

Time spent on this PR: 0.5 days

Closes #46 

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: N/A

# What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


core/src/client/types.rs
```diff
@@ -183,3 +183,16 @@ impl<T: Serialize> Serialize for Rich<T> {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TokenBalance {
+    pub contract_address: Address,
+    pub token_balance: String,
```
Eikix, line 190: I think both token_balance and error can be options,
error can remain a string I guess for now:)

i suggest token_balance to be a bytes32 (i think U256 is best suited
)

core/src/client/mod.rs
```diff
@@ -734,4 +740,72 @@ impl StarknetClient for StarknetClientImpl {
...
+    async fn get_token_balances(
+        &self,
+        address: Address,
+        contract_addresses: Vec<Address>,
+    ) -> Result<TokenBalances, KakarotClientError> {
+        let mut token_balances = Vec::new();
+        let starknet_user_address = self
+            .compute_starknet_address(address, StarknetBlockId::Tag(BlockTag::Latest))
+            .await?;
+        let entrypoint = FieldElement::from_str("balanceOf").map_err(|e| {
```
Eikix, line 766: the entrypoint should be encoded as an EVM entrypoint for kakarot execute_at_address 

`keccak256("balanceOf(address)")`

https://solidity-by-example.org/function-selector/

core/src/client/mod.rs
```diff
@@ -734,4 +740,72 @@ impl StarknetClient for StarknetClientImpl {
...
+    ///
+    /// `Ok(<TokenBalances>)` if the operation was successful.
+    /// `Err(KakarotClientError)` if the operation failed.
+    async fn get_token_balances(
+        &self,
+        address: Address,
+        contract_addresses: Vec<Address>,
+    ) -> Result<TokenBalances, KakarotClientError> {
+        let mut token_balances = Vec::new();
+        let starknet_user_address = self
```
Eikix, line 763: I wonder if we should do the conversion here,

essentially a user sends a `balanceOf()` read to Kakarot itself, meaning that Kakarot should be able to perform the conversion from EVM to Starknet address itself

Eikix: Because if at some point in Solidity someone writes:

```solidity

let myBalance = MyErc20.balanceOf(address(this))

```

address(this) is a 160-bit EVM address, but Kakarot will convert it inside its execution context, so its execute_at_address' responsibility to perform this conversion

core/src/client/mod.rs
```diff
@@ -734,4 +740,72 @@ impl StarknetClient for StarknetClientImpl {
...
+                Ok(call) => {
+                    let hex_balance = U256::from_str_radix(&call.to_string(), 16).map_err(|e| {
+                        KakarotClientError::OtherError(anyhow::anyhow!(
+                            "Failed to convert token balance to U256: {}",
+                            e
+                        ))
+                    })?;
+                    TokenBalance {
+                        contract_address: address,
+                        token_balance: Some(U256::from(hex_balance)),
```
Eikix, line 793: Isn't hex_balance already a U256?

core/src/client/mod.rs
```diff
@@ -734,4 +740,72 @@ impl StarknetClient for StarknetClientImpl {
...
+                        error: Some("".to_string()),
```
Eikix, line 794: Can be None

core/src/client/mod.rs
```diff
@@ -734,4 +740,72 @@ impl StarknetClient for StarknetClientImpl {
...
+                    }
+                }
+                Err(_) => TokenBalance {
+                    contract_address: address,
+                    token_balance: None,
+                    error: Some("Failed to get token balance".to_string()),
```
Eikix, line 800: just an idea, could be:
```
Err(e) => TokenBalance {
   contract_address: address,
   token_balance: None,
   error: Some(format!("kakarot_getTokenBalance Error:", e.to_string())
}

```

or something like that

could use also to_string_lossy or w/e works

Eikix (changes requested):
very cool PR, almost there

core/src/client/mod.rs
irisdv: Ah yes ok

core/src/client/mod.rs
```diff
@@ -1,7 +1,10 @@
+use std::str::FromStr;
```
Eikix, line 1: do you use this?

Eikix (approved):
looks good to me! is this tested?

core/src/client/mod.rs
irisdv: Yes I need it to use `from_str` in `FieldElement::from_str(&address.to_string())` line 820, or maybe there's a better way to convert a string to a FieldElement ?

Eikix: If you want, you can use

`let a = felt!("0x1939391919313");`

Eikix (commented):
Almost ready to merge

core/src/client/mod.rs
irisdv: `felt!("")` expects a string literal, is it possible to convert a variable of type `&str` or `String` into a string literal ? I couldn't find a way to do it.

Eikix: to me, string literals are &str

Eikix (approved):
lgtm

PR #90, opened 2023-02-13
[WIP] dev: add core kakarot logic tests using starknet client mock implementation

### LINKED ISSUE #86: dev: add core kakarot logic tests using starknet client mock implementation
currently we have a mock impl. that is able to return mocked values of starknet json rpc endpoint, we can leverage that to test kakarot rpc core logic and check that it returns sound evm formatted payloads

irisdv:
Time spent on this PR:

Closes #86 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Testing

Issue Number: N/A

# What is the new behavior?

Added mocks for : 
- `starknet_getTransactionByBlockIdAndIndex`
- `starknet_getTransactionReceipt`
- `starknet_getTransactionByHash`
- `starknet_call` (in the context of `get_evm_address` called in `eth_getTransactionReceipt`)

Added tests in `rpc/tests/rpc.rs` for : 
- `eth_blockNumber`
- `eth_getBlockByHash`
- `eth_getBlockByNumber`
- `eth_getBlockTransactionCountByHash`
- `eth_getBlockTransactionCountByNumber`
- `eth_getTransactionReceipt`
- `eth_getTransactionByBlockNumberAndIndex`
- `eth_getTransactionByBlockHashAndIndex`

- [ ] Yes
- [x] No

# Other information

In `wiremock_utils.rs`, added a function called `debug_entry_param` which returns the `JsonRpcRequest` entry params using different serialization implementations (`SerializeAs` and `DeserializeAs`). Wiremock needs to receive the exact parameters of a call to mock a result. This function is useful to debug and ensure entry params are the same between tests and created mocks.


core/src/client/types.rs
```diff
@@ -129,10 +129,12 @@ pub struct Header {
     /// Nonce
     pub nonce: Option<H64>,
     /// Size in bytes
+    #[serde(rename = "sizeHeader")]
```
Eikix, line 132: where is this from? 
I understand there is a name collision due to header names, but how so?

core/src/client/types.rs
```diff
@@ -129,10 +129,12 @@ pub struct Header {
...
     pub size: Option<U256>,
 
     pub mix_hash: H256,
 
+    #[serde(rename = "baseFeePerGasHeader")]
```
Eikix, line 137: same

core/src/client/types.rs
irisdv: I had an error when deserializing the response for a `Block`, which is defined like this:

```
pub struct Block {
    #[serde(flatten)]
    pub header: Header,
    pub total_difficulty: U256,
    pub uncles: Vec<H256>,
    pub transactions: BlockTransactions,
    pub size: Option<U256>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub base_fee_per_gas: Option<U256>,
}
```
The errors said there was twice the same name variable for `size` and `base_fee_per_gas`. Renaming them with serde in the Header was the only solution I could find. Is there another way to handle this kind of issues ?

MartianGreed (commented):
LGTM 🙈 !

Very tedious one ! But you're on the right path. This will ensure painless and stressless deployments..

Although some attention points : 

-  Code is becoming more and more verbose, we should factorize some components again 
    -  Wiremock instanciation in wiremock_utils can be simpler (core/src/utils/wiremock_utils.rs)
    - Tests in rpc/tests/rpc.rs are good but can be factorized too `assert_block`

I'm available to discus about it !

core/src/client/types.rs
Eikix: How come this does not happen when one requests a block through the RPC?

Eikix: Renaming it here means it is renamed for the whole app, no? which means we break the JSON API spec? Or does it mean we rename the fields if and only if we're deserializing them (meaning when serializing, we don't rename it)?

Eikix (approved):
This is a very complete and nice PR:)! 

Let's merge it,
I wonder if in the future we should refactor test_helpers, I wonder if we're testing the right thing. We should debate about it:).

There are many ways to test our client's method calls:
- Compare the resulting Block with the "original" Starknet json payload (what you are doing)
- Compare the resulting Block with the typical Ethereum Block we would expect when calling a Kakarot method

core/src/client/types.rs
irisdv: Discussing this with @MartianGreed,`#[serde(flatten)]` inline all variables from the `Block` which results in having `size` and `base_fee_per_gas` twice. Removing flatten solves the issue. Where is flatten needed ?

PR #100, opened 2023-02-17
feat: add dockerfile to kakarot-rpc

Eikix:
Adding Docker to our project in order to facilitate Paas deployment


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Build the application
+RUN cargo build --all --release
+
+# Create a new container from scratch to reduce image size
+FROM debian:buster-slim
+
+# Install any necessary dependencies
+RUN apt-get update && apt-get install -y libssl-dev ca-certificates && rm -rf /var/lib/apt/lists/*
+
+# RUN update-ca-certificates
```
ClementWalter, line 18: remove?

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+WORKDIR /usr/src/app
+
+# Copy the built binary from the previous container
+COPY --from=builder /usr/src/rpc/target/release/kakarot-rpc .
+
+# Expose the port that the RPC server will run on
+EXPOSE 3030
+
+# Run the binary
+CMD ["./kakarot-rpc"]
```
ClementWalter, line 30: though this is fine, there is apparently a best practice consisting in using `tiny`, see for example https://github.com/eqlabs/pathfinder/blob/4ab915a830953ed6f02af907937b46cb447d9a92/Dockerfile#L120

rpc/src/lib.rs
```diff
@@ -20,7 +20,7 @@ pub async fn run_server(
     starknet_client: Box<dyn KakarotClient>,
 ) -> Result<(SocketAddr, ServerHandle), RpcError> {
     let server = ServerBuilder::default()
-        .build("127.0.0.1:3030".parse::<SocketAddr>()?)
+        .build("0.0.0.0:3030".parse::<SocketAddr>()?)
```
ClementWalter, line 23: should we use a ENV var for the port ?

Dockerfile
Eikix: ok!

rpc/src/lib.rs
Eikix: ok!

ClementWalter (approved):

ClementWalter (approved):

PR #108, opened 2023-02-20
dev: update eth_getBlockTransactionCountByHash & byNumber to return KKT txs count

### LINKED ISSUE #92: dev: update eth_getBlockTransactionCountByHash & byNumber to use starknet_getBlockTransactionCount
Currently `eth_getBlockTransactionCountByHash` and `eth_getBlockTransactionCountByNumber` functions both fetch `starknet_getBlockWithTxHashes` endpoint and then count the number of transactions in the block. When block is pending, both endpoints do not return anything. 

Both functions should use `starknet_getBlockTransactionCount` which returns the number of transactions for accepted and pending blocks.

dpinones:
The `eth_getBlockTransactionCountByHash` and `eth_getBlockTransactionCountByNumber` functions get the number of transactions per block from the `starknet_getBlockTransactionCount` function. Previously they used the `starknet_getBlockWithTxHashes` function.

Updated issue resolution: https://github.com/sayajin-labs/kakarot-rpc/pull/108#issuecomment-1444718187

Time spent on this PR: 1.5 day

Closes #92 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: N/A

# What is the new behavior?

-

- [ ] Yes
- [x] No

# Other information


dpinones:
It only remains to update the tests

Eikix (changes requested):
thanks for your PR, 
I realize the issue was not perfectly clear: currently, this PR counts the number of transactions on Starknet at a certain block,
that being said, our business logic is to return the number of Kakarot transactions, which is a subset of the number of starknet transactions at a certain block.



This PR will need to be changed altered in order to return this specific number.


Changes: 
A - Change the logic so that it counts the number of Kakarot (KKT) tx, you can draw inspiration from the `starknet_tx_into_kakarot_tx` function and `filter_starknet_into_eth_txs`, since a Starknet transaction is considered (for now) a KKT tx if and only if the class hash of the sender address corresponds to the account_proxy class hash (constant or env var) from Kakarot.

B - Could you rebase it on the new state of main 

C -  add a test for your use case?

you can draw inspiration in `rpc/tests/rpc`

1. add a wiremock route for your Starknet call get_transaction_count in `core/src/utils/wiremock_utils`
2. add a test in `rpc/tests/rpc`

dpinones:
If I understood correctly, to meet the requirements of this PR, we cannot use the `get_transaction_count` function since it only returns a number representing the transaction count, and we need detailed information on each transaction to filter out only the KKT transactions.

Therefore, I made the following changes:

1. I used the `get_block_with_txs` function to obtain all transactions in a block.
2. I used the `filter_starknet_into_eth_txs` function to filter the transactions and obtain only the KKT transactions.
3. Finally, I return the count of filtered transactions that are KKT transactions.
4. I also added test cases and performed a rebase with the current state of the main branch.
5. Additionally, I added a mock for the `get_transaction_count` function, although I did not use it.

Please let me know if there are any corrections that need to be made.

core/src/utils/wiremock_utils.rs
```diff
@@ -126,6 +135,46 @@ pub async fn setup_wiremock() -> String {
...
+    // block_transaction_count latest
+    let latest_block = StarknetBlockId::Tag(BlockTag::Latest);
+    Mock::given(method("POST"))
+        .and(body_json(StarknetRpcBaseData::get_block_transaction_count(
+            [&latest_block],
+        )))
+        .respond_with(
+            ResponseTemplate::new(200)
+                .set_body_raw(
+                    include_str!("data/blocks/starknet_getBlockTransactionCount.json"),
```
Eikix, line 169: did you add this file?

since it's not a `starknet_getBlockTransactionCount` call we're doing inside our core logic, we don't have to call it that, we don't even have to mock it,

what we're really mocking is `client.get_block_with_txs` from [this line](https://github.com/sayajin-labs/kakarot-rpc/pull/108/files#diff-e6d20c367d720630c43945964dc7fdb3105ee5c321f359ce26f51c44b820f68aR449)

which means we have to "format" well a call to get_block_with_txs if it's not already being mocked properly,
if tests pass etc. maybe an already existing get_block_with_txs mock is enough for our use case of filtering txs

Eikix (changes requested):
very cool PR, thanks for being reactive and understanding,
almost ready to merge 🔥

core/src/utils/wiremock_utils.rs
dpinones: You were right, the mock was eliminated because it was not being used.

core/src/utils/wiremock_utils.rs
```diff
@@ -44,6 +44,15 @@ impl<'a, StarknetParams> StarknetRpcBaseData<'a, StarknetParams> {
         }
     }
 
+    pub fn get_block_transaction_count(params: StarknetParams) -> Self {
```
Eikix, line 47: do you still need this?

dpinones: You were right, I don't need it anymore

docs/methods/eth_getBlockTransactionCountByNumber.md
```diff
@@ -30,7 +30,7 @@ transactions in a block matching the given block number.
 
 ### Starknet methods
 
-- [starknet_getBlockWithTxs](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L44)
+- [starknet_getBlockTransactionCount](https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L373)
```
Eikix, line 33: I think we're not using it in the end

rpc/tests/rpc.rs
```diff
@@ -224,15 +224,14 @@ mod tests {
     async fn test_block_transaction_count_by_hash_is_ok() {
         let kakarot_rpc = setup_kakarot_eth_rpc().await;
         let hash =
-            H256::from_str("0x0197be2810df6b5eedd5d9e468b200d0b845b642b81a44755e19047f08cc8c6e")
+            H256::from_str("0x0449aa33ad836b65b10fa60082de99e24ac876ee2fd93e723a99190a530af0a9")
```
Eikix, line 227: why are we changing the hash btw? what does it bring?

dpinones: Because now I am using the "get_block_with_txs" method and in the wiremock_utils file it has that [hash](https://github.com/sayajin-labs/kakarot-rpc/blob/abfb3d5564b930782a153fe4bfdb770dc59cd8a6/core/src/utils/wiremock_utils.rs#L131)

Eikix (approved):
lgtm 🔥🔥🔥🔥🔥🔥🔥

PR #117, opened 2023-02-27
feat: refactor async for loops using stream

### LINKED ISSUE #98: dev: refactor async for loops and try_join_all into join_all
Currently we have the need to loop over futures.

We should use the idiomatic `join_all` method on futures instead of for loops.

`try_join_all` is suboptimal for our use case as it returns a `Result<Vec<_>>` meaning that it requires all the calls to succeed.

`join_all` returns a `Vec<Result<_>>` which lets you handle every case in a singular way.

Task: replace all For loops on Async Futures by `futures::join_all` and handle properly error cases for each vector element.

hurrikaanig:
Time spent on this PR: 1d

Closes #98

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Using for loop on async not allowing executing them asynchronously

# What is the new behavior?

Using futures::stream allowing asynchronous execution of async

- [ ] Yes
- [X] No

# Other information

I was supposed to use join_all but during my research I found that streams were better suited for this as it allow asynchronous execution. 100% open to discussion though, I don't have a deep knowledge of rust yet

### DIFF
--- a/core/src/client/mod.rs
+++ b/core/src/client/mod.rs
@@ -1,6 +1,7 @@
 use std::str::FromStr;
 
 use eyre::Result;
+use futures::future::join_all;
 use jsonrpsee::types::error::CallError;
 use std::convert::From;
 
@@ -841,7 +842,6 @@ impl KakarotClient for KakarotClientImpl {
         address: Address,
         contract_addresses: Vec<Address>,
     ) -> Result<TokenBalances, KakarotClientError> {
-        let mut token_balances = Vec::new();
         let entrypoint = hash_to_field_element(keccak256("balanceOf(address)")).map_err(|e| {
             KakarotClientError::OtherError(anyhow::anyhow!(
                 "Failed to convert entrypoint to FieldElement: {}",
@@ -854,25 +854,28 @@ impl KakarotClient for KakarotClientImpl {
                 e
             ))
         })?;
-
-        for token_address in contract_addresses {
+        let handles = contract_addresses.into_iter().map(|token_address| {
             let calldata = vec![entrypoint, felt_address];
-            let call = self
-                .call_view(
-                    token_address,
-                    Bytes::from(vec_felt_to_bytes(calldata).0),
-                    StarknetBlockId::Tag(BlockTag::Latest),
-                )
-                .await;
 
-            let token_balance = match call {
+            self.call_view(
+                token_address,
+                Bytes::from(vec_felt_to_bytes(calldata).0),
+                StarknetBlockId::Tag(BlockTag::Latest),
+            )
+        });
+        let token_balances = join_all(handles)
+            .await
+            .into_iter()
+            .map(|token_address| match token_address {
                 Ok(call) => {
-                    let hex_balance = U256::from_str_radix(&call.to_string(), 16).map_err(|e| {
-                        KakarotClientError::OtherError(anyhow::anyhow!(
-                            "Failed to convert token balance to U256: {}",
-                            e
-                        ))
-                    })?;
+                    let hex_balance = U256::from_str_radix(&call.to_string(), 16)
+                        .map_err(|e| {
+                            KakarotClientError::OtherError(anyhow::anyhow!(
+                                "Failed to convert token balance to U256: {}",
+                                e
+                            ))
+                        })
+                        .unwrap();
                     TokenBalance {
                         contract_address: address,
                         token_balance: Some(hex_balance),
@@ -884,9 +887,8 @@ impl KakarotClient for KakarotClientImpl {
                     token_balance: None,
                     error: Some(format!("kakarot_getTokenBalances Error: {e}")),
                 },
-            };
-            token_balances.push(token_balance);
-        }
+            })
+            .collect();
 
         Ok(TokenBalances {
             address,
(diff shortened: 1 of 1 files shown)


hurrikaanig:
Since we wanted to iterate over async functions and from what I understood streams were made for this I thought it could be a more idiomatic way of doing it

Eikix:
> Since we wanted to iterate over async functions and from what I understood streams were made for this I thought it could be a more idiomatic way of doing it

Thanks a lot for your input, I asked the help of @tdelabro, a great rustacean, who'll help us understand what is the most rustic way of handling this case:)

core/src/client/mod.rs
```diff
@@ -1299,15 +1304,20 @@ impl KakarotClient for KakarotClientImpl {
...
+                    .await
+            })
+            .filter_map(|transaction| async {
+                match transaction {
+                    Ok(tx) => Some(tx),
+                    Err(_) => None,
+                }
+            })
+            .collect()
+            .await;
```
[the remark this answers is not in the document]
tdelabro: The core problem with what you are doing is that, although you are using `async` functions (here `stream`), your code is still processed iteratively.

`stream::iter` will yield the items it contains one after the other, exactly like a standard iterator. The main difference is that, if the next item in the stream (which is a future) is not ready yet, the program will wait for it in a non-blocking manner. But it won't go for the item following one until it finished processing the first one.

In your case, there will be no waiting at all because all the items are already available. They all come from a `Vec` stored somewhere in your program memory, so they are immediately available. `stream::iter()` will here behave exactly like a regular `Vec::iter`.

I understand the reason why you want to use it is to be able to use `then`, an `async` function which is not available in the regular `Vec::iter` API. But you actually don't need it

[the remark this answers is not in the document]
tdelabro: What I propose instead is:
1) Create all the `Futures` you intend to execute. (more exactly an iterator over all those `Futures`)
```rust
let handles = initial_transactions
            .into_iter()
            .map(|starknet_transaction| {
                self.starknet_tx_into_kakarot_tx(starknet_transaction, blockhash_opt, blocknum_opt)
            });
```
The trick is to not call `await` on `self.starknet_tx_into_kakarot_tx(starknet_transaction, blockhash_opt, blocknum_opt)`, this way nothing it executed at this point.

2) call [futures::future::join_all](https://docs.rs/futures/latest/futures/future/fn.join_all.html) on your iterator
```rust
futures::future::join_all(handles).await
```
`join_all` will transform all your `Futures` into one single `Future` (the one you await), that will internally wait for all your futures to resolve, but concurrently. Meaning that all the `I/O` requests will be fired more or less "at the same time" and the responses waited at once.

It will yield a `Vec<Result<Transaction, KakarotClientError>>` that we can then normally `iter` and `filter`.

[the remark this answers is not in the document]
tdelabro: I propose you follow the same implementation for the other two functions: `token_balances` and `transaction_receipt` once @Eikix has confirmed that all those futures are safe to be executed concurrently.

tdelabro (changes requested):

core/src/client/mod.rs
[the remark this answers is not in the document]
hurrikaanig: Ok thanks a lot I understand way better how it works now. But does it mean that the example given [here](https://docs.rs/futures/latest/futures/stream/fn.iter.html) is not really accurate since the vec in the stream is composed of already available items ? I just want to make sure I get it right.

core/src/client/mod.rs
```diff
@@ -1310,15 +1312,17 @@ impl KakarotClient for KakarotClientImpl {
...
-        let mut transactions_vec = vec![];
-        for transaction in initial_transactions {
-            let tx_value = self
-                .starknet_tx_into_kakarot_tx(transaction, blockhash_opt, blocknum_opt)
-                .await;
-            if let Ok(val) = tx_value {
-                transactions_vec.push(val)
-            }
-        }
+        // TODO replace filter_map to handle errors
```
Eikix, line 1315: why is there a TODO here?

Eikix (commented):
v cool PR! 🔥

core/src/client/mod.rs
hurrikaanig: Errors were not handled previously but maybe we should I'm not sure

Eikix: I think they're handled in the filter_map, no?

Eikix: Ahh okay I misunderstood, you said, to replace the filter_map with actual error handling!
I think we don't need to do that, we might want to refactor the fact that `starknet_tx_into_kakarot_tx` returns a result instead of an option (would be better imo)

Let's remove this comment and ready to merge:)!

hurrikaanig: I'm only using filter_map to remove them from the vector as it was done here: https://github.com/sayajin-labs/kakarot-rpc/blob/ab59ebfb3878b6606e293f7ce42150727cc973bb/core/src/client/mod.rs#L1318

Errors were not pushed in transactions_vec. I reproduced this behavior but I'm not sure if that's what we should do

Eikix: imho it's fine:)!

hurrikaanig: Removed the comment !

Eikix (approved):
lgtm 🔥

PR #128, opened 2023-03-03
[WIP] dev: update bytes_to_felt_vec

### LINKED ISSUE #84: dev: add tests for `vec_felt_to_bytes`
Add unit testing for the function `vec_felt_to_bytes` in the helpers.rs file

dpinones:
Add unit testing for the function vec_felt_to_bytes in the helpers.rs file

Closes #84 

# Pull Request type 

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: N/A

# What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

# Other information

### DIFF
--- a/core/src/helpers.rs
+++ b/core/src/helpers.rs
@@ -260,7 +260,14 @@ pub fn hash_to_field_element(hash: H256) -> Result<FieldElement, KakarotClientEr
 }
 
 pub fn bytes_to_felt_vec(bytes: Bytes) -> Vec<FieldElement> {
-    bytes.to_vec().into_iter().map(FieldElement::from).collect()
+    let bytes_vec = bytes.to_vec();
+    let mut result = Vec::new();
+    for i in 0..bytes_vec.len() / 32 {
+        let mut buf = [0; 32];
+        buf.copy_from_slice(&bytes_vec[i * 32..(i + 1) * 32]);
+        result.push(FieldElement::from_bytes_be(&buf).unwrap());
+    }
+    result
 }
 
 /// Author: https://github.com/xJonathanLEI/starknet-rs/blob/447182a90839a3e4f096a01afe75ef474186d911/starknet-accounts/src/account/execution.rs#L166
@@ -322,6 +329,17 @@ mod tests {
                 FieldElement::from(10_u64)
             ]
         );
+    }
+
+    #[test]
+    fn test_bytes_to_felt_vec_and_vec_felt_to_bytes() {
+        let input = Bytes::from(vec![
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 1,
+        ]);
+        let expected_output = vec![FieldElement::from(1_u64)];
+        assert_eq!(bytes_to_felt_vec(input.clone()), expected_output);
+        assert_eq!(vec_felt_to_bytes(expected_output), input);
     }
 
     #[test]


dpinones:
Danilo [commented](https://github.com/sayajin-labs/kakarot-rpc/issues/84#issuecomment-1453056904)  that the [`vec_felt_to_bytes`](https://github.com/sayajin-labs/kakarot-rpc/blob/ab59ebfb3878b6606e293f7ce42150727cc973bb/core/src/helpers.rs#L210) and [`bytes_to_felt_vec` ](https://github.com/sayajin-labs/kakarot-rpc/blob/ab59ebfb3878b6606e293f7ce42150727cc973bb/core/src/helpers.rs#L262) methods were inverse to each other. However, after running a test, it was found that they were not.

```
#[test]
fn test_bytes_to_felt_vec_and_vec_felt_to_bytes() {
    let input = Bytes::from(vec![
        1
    ]);

    let expected_output = vec![
        FieldElement::from(1_u64),
    ];
    assert_eq!(bytes_to_felt_vec(input.clone()), expected_output);
    assert_eq!(vec_felt_to_bytes(expected_output), input);
}
```

Analyzing the code, it was concluded that the `vec_felt_to_bytes` method was working correctly, but the `bytes_to_felt_vec` method had a problem. Instead of reading blocks of 32 bytes, it was reading blocks of 8 bytes.

I propose to correct the `bytes_to_felt_vec` method as follows(in the comment below)

If my analysis is correct, it is necessary to update all calls to the `bytes_to_felt_vec` method and add more test cases for both methods.
Please let me know if there are any corrections that need to be made. Thanks :)

core/src/helpers.rs
```diff
@@ -260,7 +260,14 @@ pub fn hash_to_field_element(hash: H256) -> Result<FieldElement, KakarotClientEr
 }
 
 pub fn bytes_to_felt_vec(bytes: Bytes) -> Vec<FieldElement> {
-    bytes.to_vec().into_iter().map(FieldElement::from).collect()
+    let bytes_vec = bytes.to_vec();
```
Eikix, line 263: why are we doing this refactor?
here, our goal is doing 1 byte = 1 felt 
Bytes -> Vec<u8> -> Vec<FieldElement>

core/src/helpers.rs
```diff
@@ -324,6 +331,17 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_bytes_to_felt_vec_and_vec_felt_to_bytes() {
+        let input = Bytes::from(vec![
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 1,
+        ]);
+        let expected_output = vec![FieldElement::from(1_u64)];
```
Eikix, line 340: here we'd want to do Bytes(0,0,0,0,0,...,1) -> vec![0,0,0,0...,1] -> vec![felt!(0), felt!(0), ..., felt!(1)]

Eikix (changes requested):
nice PR! I think there was a misunderstanding in what we expected the function to accomplish

core/src/helpers.rs
Eikix: Yes, but here, since we are dealing with bytes, we know the value will be between 0 and 255 (since bytes.to_vec() will yield a Vec<u8>)

Eikix:
I think there was a misunderstanding on this PR / issue. I apologize for this.

Maybe we spoke too fast when we said both of these functions inverted each other, or our language was not precise enough.


In my view, @danilowhk can confirm, but `vec_felt_to_bytes` represents a vector of field elements that are between 0 and 255. In this util function, there should be some kind of check such that we know the Vec<FieldElement> is in fact a Vec<u8>.

Our sole purpose is to switch between types, but the underlying data will always be bytes, i.e. single values between 0 and 255. 

`vec_felt_to_bytes`'s flow:  Vec<FieldElement> -> Vec<u8> -> Bytes (granted each individual FieldElement is <2**8)
'bytes_to_vec_felt`'s flow: Bytes -> Vec<u8> -> Vec<FieldElement>

PR #167, opened 2023-06-21
refactor: extract out all constants being used as hardcoded values or to mock realdata in a single file 🚀 

### LINKED ISSUE #154: refactor: create constants for hardcoded values shared in kakarot core
Currently we use a lot of hard coded values in various places in `mod.rs` of `core` crate. We need to replace them with constants, that sit in constant.rs. 

Example:

```rust

async fn starknet_block_to_eth_block(
        &self,
        block: MaybePendingStarknetBlock,
    ) -> Result<RichBlock, KakarotClientError> {
        // Fixed fields in the Ethereum block as Starknet does not have these fields

        //TODO: Fetch real data
        let gas_limit = U256::from(1_000_000);

        //TODO: Fetch real data
        let gas_used = U256::from(500_000);

        //TODO: Fetch real data
        let difficulty = U256::ZERO;

        //TODO: Fetch real data
        let nonce: Option<H64> = Some(H64::zero());

        //TODO: Fetch real data
        let size: Option<U256> = Some(U256::from(1_000_000));

        // Bloom is a byte array of length 256
        let logs_bloom = Bloom::default();
        let extra_data = Bytes::from(b"0x00");
        //TODO: Fetch real data
        let total_difficulty: U256 = U256::ZERO;
        //TODO: Fetch real data
        let base_fee_per_gas = self.base_fee_per_gas();
        //TODO: Fetch real data
        let mix_hash = H256::zero();

// ...rest of code
```

bajpai244:
There are multiple places where we are using hardcoded values to mock real data, the pr assembles all such constants in a single `mock_evm_data` module, inside `constants.rs`.

The name of the module comes from the fact that we are using these constants in place of mock values.

As we keep changing these values with real data, we can keep removing these constants as well from the module + now change in their values will also be atomic.

Closes #154 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Currently, we are using hardcoded values in place of constants, whereas we extract them to a single source so that a change to their values is atomic.

Issue Number: #154 

# What is the new behavior?

- All mocked hardcoded values have been extracted to constants.

- [ ] Yes
- [X] No

# Other information

The heuristic used to find these values were:
- Search for `0_0` pattern throughout the codebase
- Look for TODO comments stating the value to be replaced with real value in future.

### DIFF
--- a/crates/core/src/client/constants.rs
+++ b/crates/core/src/client/constants.rs
@@ -1,3 +1,7 @@
+use lazy_static::lazy_static;
+use reth_primitives::{H256, H64, U128, U256, U8};
+use starknet::core::types::FieldElement;
+
 pub const CHAIN_ID: u64 = 1_263_227_476;
 
 pub const STARKNET_NATIVE_TOKEN: &str = "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7";
@@ -37,4 +41,21 @@ pub mod gas {
     /// on a "first come first served" basis by the Sequencer.
     /// As a result, the priority fee is set to 0.
     pub const MAX_PRIORITY_FEE_PER_GAS: U128 = U128::ZERO;
+}
+
+// This module contains constants which are being used in place of real data that should be fetched
+// in production.
+lazy_static! {
+    pub static ref GAS_LIMIT: U256 = U256::from(1_000_000u64);
+    pub static ref GAS_USED: U256 = U256::from(500_000u64);
+    pub static ref CUMULATIVE_GAS_USED: U256 = U256::from(1_000_000u64);
+    pub static ref EFFECTIVE_GAS_PRICE: U128 = U128::from(1_000_000u64);
+    pub static ref SIZE: Option<U256> = Some(U256::from(1_000_000u64));
+    pub static ref MAX_FEE: FieldElement = FieldElement::from(100_000_000_000_000_000u64);
+    pub static ref ESTIMATE_GAS: U256 = U256::from(100_000_000_000_000_000u64);
+    pub static ref TRANSACTION_TYPE: U8 = U8::from(0);
+    pub static ref NONCE: Option<H64> = Some(H64::zero());
+    pub static ref MIX_HASH: H256 = H256::zero();
+    pub static ref DIFFICULTY: U256 = U256::from(0);
+    pub static ref TOTAL_DIFFICULTY: Option<U256> = None;
 }
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -45,7 +45,7 @@ use constants::selectors::BYTECODE;
 use self::client_api::{KakarotClient, KakarotClientError};
 use self::constants::gas::{BASE_FEE_PER_GAS, MAX_PRIORITY_FEE_PER_GAS};
 use self::constants::selectors::{BALANCE_OF, COMPUTE_STARKNET_ADDRESS, GET_EVM_ADDRESS};
-use self::constants::STARKNET_NATIVE_TOKEN;
+use self::constants::{MAX_FEE, STARKNET_NATIVE_TOKEN};
 use crate::models::convertible::ConvertibleStarknetBlock;
 use crate::models::{BlockWithTxHashes, BlockWithTxs, TokenBalance, TokenBalances};
 
@@ -862,7 +862,7 @@ impl KakarotClient for KakarotClientImpl<JsonRpcClient<HttpTransport>> {
         let calldata = raw_starknet_calldata(self.kakarot_address, bytes);
 
         // Get estimated_fee from Starknet
-        let max_fee = FieldElement::from(100_000_000_000_000_000_u64);
+        let max_fee = *MAX_FEE;
 
         let signature = vec![];
 
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -29,3 +29,4 @@ serde = { version = "1.0" }
 futures = "0.3.26"
 
 wiremock = "0.5.17"
+lazy_static = "1.4.0"
--- a/crates/core/src/client/helpers.rs
+++ b/crates/core/src/client/helpers.rs
@@ -1,7 +1,6 @@
 use eyre::Result;
 use reth_primitives::{
-    Address, BlockId as EthBlockId, BlockNumberOrTag, Bloom, Bytes, Signature, TransactionSigned, H160, H256, U128,
-    U256, U8,
+    Address, BlockId as EthBlockId, BlockNumberOrTag, Bloom, Bytes, Signature, TransactionSigned, H160, H256, U256,
 };
 use reth_rlp::Decodable;
 use reth_rpc_types::TransactionReceipt;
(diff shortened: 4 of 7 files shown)


crates/core/src/client/constants.rs
```diff
@@ -39,3 +39,22 @@ pub mod gas {
...
+    pub const GAS_USED: u64 = 500_000;
+    pub const CUMULATIVE_GAS_USED: u64 = 1_000_000;
+    pub const EFFECTIVE_GAS_PRICE: u64 = 1_000_000;
+    pub const SIZE: u64 = 1_000_000;
+    pub const MAX_FEE: u64 = 100_000_000_000_000_000;
+    pub const ESTIMATE_GAS: u64 = 100_000_000_000_000_000;
+    pub const TRANSACTION_TYPE: u64 = 0;
+    pub const NONCE: H64 = H64::zero();
+    pub const MIX_HASH: H256 = H256::zero();
+    pub const DIFFCIULTY: u64 = 0;
```
Eikix, line 58: typo

crates/core/src/client/constants.rs
```diff
@@ -39,3 +39,22 @@ pub mod gas {
...
+    pub const TOTAL_DIFFCIULTY: u64 = 0;
```
Eikix, line 59: typo,
thinking of setting it as None, always. wdyt

crates/core/src/client/constants.rs
```diff
@@ -39,3 +39,22 @@ pub mod gas {
     /// As a result, the priority fee is set to 0.
     pub const MAX_PRIORITY_FEE_PER_GAS: U128 = U128::ZERO;
 }
+
+// This module contains constants which are being used in place of real data that should be fetched
+// in production.
+pub mod mock_evm_data {
```
Eikix, line 45: For reth types, such as `U256::from(...)` etc.
Can we try to do an implem with lazy_static! macro?

Then, the constants could be directly typed as U256 and not have to be derived later?

crates/core/src/client/constants.rs
```diff
@@ -39,3 +39,22 @@ pub mod gas {
...
+
+    pub const GAS_LIMIT: u64 = 1_000_000;
+    pub const GAS_USED: u64 = 500_000;
+    pub const CUMULATIVE_GAS_USED: u64 = 1_000_000;
+    pub const EFFECTIVE_GAS_PRICE: u64 = 1_000_000;
+    pub const SIZE: u64 = 1_000_000;
+    pub const MAX_FEE: u64 = 100_000_000_000_000_000;
+    pub const ESTIMATE_GAS: u64 = 100_000_000_000_000_000;
+    pub const TRANSACTION_TYPE: u64 = 0;
+    pub const NONCE: H64 = H64::zero();
```
Eikix, line 56: thinking out loud, maybe this should be None as well, always.

crates/core/src/client/mod.rs
```diff
@@ -944,29 +947,29 @@ impl KakarotClient for KakarotClientImpl {
...
-        let difficulty = U256::ZERO;
+        let difficulty = U256::from(DIFFCIULTY);
 
         //TODO: Fetch real data
-        let nonce: Option<H64> = Some(H64::zero());
+        let nonce: Option<H64> = Some(NONCE);
 
         //TODO: Fetch real data
-        let size: Option<U256> = Some(U256::from(1_000_000));
+        let size: Option<U256> = Some(U256::from(SIZE));
```
Eikix, line 962: Size should also always be None imho
I'm thinking out loud and this is kinda out of the scope of this PR:)!

But basically, some fields in the Eth block/Eth Tx/ Eth Transaction are optional and we're putting in sentinel values instead of just None, even though we have no intention of ever filling them.


TL;DR->
Some values have hardcoded values that we'll one day replace with live data,
Some values are optional and should always be None imho

wdyt

crates/core/src/client/constants.rs
bajpai244: fix [here](https://github.com/kkrt-labs/kakarot-rpc/pull/167/commits/6d284612301da01fcc5ea39f9e98949f41b7cd55).

crates/core/src/client/constants.rs
bajpai244: I guess you are talking about `DIFFICULTY` and `TOTAL_DIFFICULTY` right?

It makes sense for them to be none, although I would say, we should make a separate issue for that, and fix it in a different PR, it will require changing a little bit in tests as well { we test for Some(U256::ZERO) }.

crates/core/src/client/constants.rs
bajpai244: Are you talking about `TRANSACTION_TYPE`?

If that is the case, then can you expand on why it should be None?

bajpai244: Oh! Gotcha, you are talking about `NONCE`, yes, makes sense!

crates/core/src/client/mod.rs
bajpai244: Can you expand on why `size` should be None? according to the spec it is supposed to be part of the response body + why will its value be None if it measures the block size?

crates/core/src/client/constants.rs
bajpai244: Lemme investigate this one!

crates/core/src/client/constants.rs
Eikix: Fine with me, can you create an issue?

crates/core/src/client/constants.rs
Eikix: yes Nonce,
I guess it can be in the same issue as Total_Difficulty then

crates/core/src/client/mod.rs
Eikix: It's supposed to be the block size is bytes,
I guess we can put is as None for now and add a todo to dynamically "size" a block -> it's simply that it's unclear how we size a KKT block for now

I guess our two options are:
- once we get a starknet block, we convert it to be an eth (kakarot) block, and then size the resulting struct, and then add the size field
- or we size directly the starknet block which is the 'reality' of the chain
- or we keep it as none for now, until we know its true use -> who wants / needs size? who uses it?

Eikix:
> @Eikix I guess we can boil it down to this:
> 
> * We have some data we mocking, when we get real data, we can remove those constants.
> * There are some values that after the merge don't make sense, and should always be None: Like, difficulty, nonce, etc.
> 
> We can limit this PR to extracting all constants to one place, then work on the second point of this comment in a separate issue!
> 
> wdyt?

That works for me.
Then before approving, can we investigate how doable it is to use lazy_static! instead of pure consts?
My understanding -> you cannot use reth_rpc_types in the constant.rs file as their type is not "const" compatible (correct me if im wrong), but it's not so elegant to declare them as u64 etc. and then do the conversion in the applicative code.

crates/core/src/client/constants.rs
bajpai244: Gotcha!

bajpai244:
@Eikix here is a new commit, with lazy_static! approach.

For anyone else reading outdated conversations, those commits were discarded because of complex conflicts during rebase with #159, hence a new commit was simply applied on top of it.

bajpai244:
@Eikix also opened an issue as per our discussion [here](https://github.com/kkrt-labs/kakarot-rpc/issues/175).

Eikix (approved):
great job! 🥳

Issue #146, opened 2023-06-13
feat: eth_getTransactionCount


bajpai244:
# eth_getTransactionCount

## Metadata

- name: getTransactionCount
- prefix: eth
- state: ⚠️
- [specification](https://github.com/ethereum/execution-apis/api-documentation/)

## Specification Description

Describe the method

### Parameters

<img width="1359" alt="image" src="https://github.com/kkrt-labs/kakarot-rpc/assets/41180869/a1c642cf-0e13-4ab7-9d8a-d1515e523dc7">

### Returns

<img width="810" alt="image" src="https://github.com/kkrt-labs/kakarot-rpc/assets/41180869/b607aa7a-4876-441c-a781-55a83f5595cb">

## Kakarot Logic

**todo**

### Kakarot methods

**todo**

### Starknet methods

**todo**

bajpai244:
@Eikix I guess I can take this one, can be a good start as well for me to play around more with the rpc!

PR #181, opened 2023-06-22
feat: add support for eth_getTransactionCount 🚀

bajpai244:
Implements the eth_getTransactionCount API, this will allow fetching nonce for an account and will allow easy use of tools like Metamask.

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

The mentioned endpoint is not implemented, and a fixed value 3 is returned.

Resolves: #146 

# What is the new behavior?

- the real nonce associated with a given EOA is returned

- [ ] Yes
- [x] No

### DIFF
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -642,8 +642,8 @@ impl KakarotClient for KakarotClientImpl<JsonRpcClient<HttpTransport>> {
     async fn nonce(&self, ethereum_address: Address, block_id: StarknetBlockId) -> Result<U256, KakarotClientError> {
         let starknet_address = self.compute_starknet_address(ethereum_address, &block_id).await?;
 
-        let nonce_felt = self.inner.get_nonce(block_id, starknet_address).await?;
-        let nonce = U256::from_be_bytes(nonce_felt.to_bytes_be());
+        let nonce = self.inner.get_nonce(block_id, starknet_address).await?;
+        let nonce = U256::from_be_bytes(nonce.to_bytes_be());
 
         Ok(nonce)
     }
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -633,7 +633,7 @@ impl KakarotClient for KakarotClientImpl<JsonRpcClient<HttpTransport>> {
 
     /// Get the nonce for a given ethereum address
     /// Convert Ethereum address to Starknet address
-    /// Return the nonce, by calling `get_nonce` on the account contract
+    /// Return the nonce, by calling `get_nonce` for the addresss via starknet rpc
     /// * `ethereum_address` - The EVM address to get the nonce of
     /// * `block_id` - The block to get the nonce at
     ///
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -44,7 +44,7 @@ use constants::selectors::BYTECODE;
 
 use self::client_api::{KakarotClient, KakarotClientError};
 use self::constants::gas::{BASE_FEE_PER_GAS, MAX_PRIORITY_FEE_PER_GAS};
-use self::constants::selectors::{BALANCE_OF, COMPUTE_STARKNET_ADDRESS, GET_EVM_ADDRESS};
+use self::constants::selectors::{BALANCE_OF, COMPUTE_STARKNET_ADDRESS, GET_EVM_ADDRESS, GET_NONCE};
 use self::constants::{MAX_FEE, STARKNET_NATIVE_TOKEN};
 use crate::models::convertible::ConvertibleStarknetBlock;
 use crate::models::{BlockWithTxHashes, BlockWithTxs, TokenBalance, TokenBalances};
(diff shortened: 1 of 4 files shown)


Eikix (changes requested):
I notice you're using the `account_contract.get_nonce()` by using `let nonce_felt = self.inner.call(request, block_id).await?;`

I think we should use self.inner.get_nonce(starknet_address) instead, as we might remove the get_nonce() function in the future

crates/rpc/src/eth_rpc.rs
```diff
@@ -147,7 +148,16 @@ impl EthApiServer for KakarotEthRpc {
     }
 
     async fn transaction_count(&self, _address: Address, _block_number: Option<BlockId>) -> Result<U256> {
-        Ok(U256::from(3))
+        let starknet_block_id = _block_number.map(ethers_block_id_to_starknet_block_id);
+        let starknet_block_id = match starknet_block_id {
```
Eikix, line 152: why are we doing this? it feels a bit verbose for no reason

crates/rpc/src/eth_rpc.rs
```diff
@@ -147,7 +148,16 @@ impl EthApiServer for KakarotEthRpc {
...
+            Some(Ok(b)) => b,
+            Some(Err(e)) => return Err(e.into()),
+            None => StarknetBlockId::Tag(BlockTag::Latest),
+        };
+
+        let transaction_count = self.kakarot_client.nonce(_address, starknet_block_id).await?;
```
Eikix, line 158: nice! logic is fine here, so we'll need to refactor the inner nonce() function -> don't use account_contract function, rather use starknet JSON RPC route getNonce()

bajpai244:
> I notice you're using the `account_contract.get_nonce()` by using `let nonce_felt = self.inner.call(request, block_id).await?;`
> 
> I think we should use self.inner.get_nonce(starknet_address) instead, as we might remove the get_nonce() function in the future

@Eikix, are you trying to say, that I should create a new function `get_nonce()`, which should contain this logic, and abstract how we get the nonce? So that all functions inside the client, that want nonce, use the same function `get_nonce()`, and we just need to make one atomic change in the future, if there is a change at the protocol level?

image

crates/rpc/src/eth_rpc.rs
bajpai244: We are doing this because the BlockID we get in the arguments of this function are from reth, and we have to convert them to StarknetBlockID types.

We also have to handle the default case of using the latest if the BlockID is `None`.

Is there are some other way you think we can simply this?

crates/rpc/src/eth_rpc.rs
bajpai244: But, we discussed on the chat right, that we should use the protocol level nonce, which is from the account contract?

Will the nonce of the Account contract deployed for the EOA by Kakarot always be the same as the nonce of the contract in Starknet?

If that is the case, then it makes sense to move to Starknet RPC implementation.

Eikix: > we discussed on the chat, that we should use the protocol level nonce
as far as i understand, this protocol level nonce is called using Starknet RPC client, not Account.get_nonce() 
Account.get_nonce() is a reproduction of the protocol level nonce.

crates/rpc/src/eth_rpc.rs
Eikix: alright, we can keep it this way for now, as we'll create a wrapper type later!

crates/rpc/src/eth_rpc.rs
bajpai244: Gotcha, as discussed, I have updated to use starknet protocol level nonces.

crates/core/src/client/mod.rs
```diff
@@ -631,6 +631,23 @@ impl KakarotClient for KakarotClientImpl<JsonRpcClient<HttpTransport>> {
         Ok(Address::from(evm_address_sliced))
     }
 
+    /// Get the nonce for a given ethereum address
+    /// Convert Ethereum address to Starknet address
+    /// Return the nonce, by calling `get_nonce` on the account contract
```
Eikix, line 636: Update comments to reflect new refactor and lgtm

bajpai244: Ah, nice catch!

Done [here](https://github.com/kkrt-labs/kakarot-rpc/pull/181/commits/a370e83eb4344dbd77b2b2b1591ea8599945302e).

crates/core/src/client/mod.rs
```diff
@@ -631,6 +631,23 @@ impl KakarotClient for KakarotClientImpl<JsonRpcClient<HttpTransport>> {
...
+    /// Return the nonce, by calling `get_nonce` for the addresss via starknet rpc
+    /// * `ethereum_address` - The EVM address to get the nonce of
+    /// * `block_id` - The block to get the nonce at
+    ///
+    /// ### Returns
+    /// * `Result<U256, KakarotClientError>` - The nonce of the EVM address
+    async fn nonce(&self, ethereum_address: Address, block_id: StarknetBlockId) -> Result<U256, KakarotClientError> {
+        let starknet_address = self.compute_starknet_address(ethereum_address, &block_id).await?;
+
+        let nonce_felt = self.inner.get_nonce(block_id, starknet_address).await?;
```
ClementWalter, line 645: no need to name with type suffixed. Rust encourages to do 

```
let nonce = self.inner ...
let nonce = U256::from_be_bytes(nonce.to_bytes_be());
```

crates/core/src/client/mod.rs
```diff
@@ -631,6 +631,23 @@ impl KakarotClient for KakarotClientImpl<JsonRpcClient<HttpTransport>> {
...
+        let nonce = U256::from_be_bytes(nonce_felt.to_bytes_be());
```
ClementWalter, line 646: can't we make `nonce.into()` ?

crates/rpc/src/eth_rpc.rs
```diff
@@ -147,7 +148,16 @@ impl EthApiServer for KakarotEthRpc {
     }
 
     async fn transaction_count(&self, _address: Address, _block_number: Option<BlockId>) -> Result<U256> {
-        Ok(U256::from(3))
```
ClementWalter, line 150: 😂

crates/core/src/client/mod.rs
bajpai244: Yeah. I know shadowing is allowed, I guess a type suffix just makes it clear to read, like:
- Oh, yes it was a felt { without going to definition of `compute_starknet_address` }
- This is why we are now converting to a U256

I know code editors can show the type for upper `nonce`, but in absence of them, it makes it more readable.

What say?

ClementWalter: I'd remove this as types as supposed to be obvious, and the rust book says not to do this naming thing, so let's stay close to it

bajpai244: Gotcha!

crates/core/src/client/mod.rs
bajpai244: @greged93 has a [PR](https://github.com/kkrt-labs/kakarot-rpc/pull/180) for this.

So, we have decided, that we can merge this, then he can rebase and implement this for this conversion as well.

crates/rpc/src/eth_rpc.rs
bajpai244: 😆

crates/core/src/client/mod.rs
bajpai244: done [here](https://github.com/kkrt-labs/kakarot-rpc/pull/181/commits/cf46bdd8e5adc038658cf7783c2a8bef1f9f8311).

Eikix (approved):
lgtm 🔥

Issue #199, opened 2023-06-23
dev: refactor environment variables, and do environment validation before RPC start


bajpai244:
tldr:
- before startup of rpc make sure essential environment variables are there and not empty { will act as a great smoke test }
- group parsing of all these variables in a single file, where they are exposed as constants, but their value is provided at runtime ( might be prefixed with `ENV` to specify these are environment variables and not normal constants ).

This will improve the readability and quality of the code, as we would be doing environment validation, and variable parsing only once, and can then use this variable like how we use constants. We can make use of lazy_static! strategy.

bajpai244:
working on this one!

@Eikix you can assign me for this.

PR #201, opened 2023-06-26
refactor: refactor environment variable handling

bajpai244:
Refactor environment variable parsing to a single file, and move validation to a single function.

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

As of now every time we need an environment variable, we parse it and then do a `map_err`,  in the future we might need more environment variables, and they might be used in various files { they are already being used in 2 } + some of these variables might only be needed when a special branch of the code is executed, and if not present might panic the service later.

This PR move parsing of all these variables and exposes them as lazy statics, it also provides a validation function, which can be run before the start of the service, if some critical variable is not present, it will panic the service before it even starts. Hence we can also use this function as a smoke test. i.e. making sure that we don't we do tests until this function passes.

The PR is more for our future needs, where the RPC might become more and more configurable, and for those configurations, we might need more and more environment variables.

Resolves: #199 

# What is the new behavior?

- All variables are exposed as lazy_statics 
- An environment validation function takes care that the service doesn't start before all environment variables are present
- All environment variables have been pre_fixed with `ENV` making sure that to reader it is clear that they are environment variables

- [ ] Yes
- [x] No

# Other information

As of now there is only test, which is for the success case of the validation function, we can investigate in future on how to test panic cases in the same test, as panics test pass when they are run individually, but tend to fail when run with other tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,3 +28,4 @@ reth-rlp = { git = "https://github.com/paradigmxyz/reth.git", rev = "fb710e5" }
 reth-primitives = { git = "https://github.com/paradigmxyz/reth.git", rev = "fb710e5" }
 async-trait = "0.1.58"
 jsonrpsee = { version = "0.18.2", features = ["full"] }
+lazy_static = "1.4.0"
--- a/crates/core/src/client/config.rs
+++ b/crates/core/src/client/config.rs
@@ -1,0 +1,40 @@
+use eyre::Result;
+use starknet::core::types::FieldElement;
+
+use super::errors::ConfigError;
+
+fn get_env_var(name: &str) -> Result<String, ConfigError> {
+    std::env::var(name).map_err(|_| ConfigError::EnvironmentVariableMissing(name.into()))
+}
+
+pub struct StarknetConfig {
+    pub starknet_rpc: String,
+    pub kakarot_address: FieldElement,
+    pub proxy_account_class_hash: FieldElement,
+}
+
+impl StarknetConfig {
+    pub fn new(starknet_rpc: &str, kakarot_address: FieldElement, proxy_account_class_hash: FieldElement) -> Self {
+        StarknetConfig { starknet_rpc: String::from(starknet_rpc), kakarot_address, proxy_account_class_hash }
+    }
+
+    pub fn from_env() -> Result<Self, ConfigError> {
+        let starknet_rpc_url = get_env_var("STARKNET_RPC_URL")?;
+
+        let kakarot_address = get_env_var("KAKAROT_ADDRESS")?;
+        let kakarot_address = FieldElement::from_hex_be(&kakarot_address).map_err(|_| {
+            ConfigError::EnvironmentVariableSetWrong(format!(
+                "KAKAROT_ADDRESS should be provided as a hex string, got {kakarot_address}"
+            ))
+        })?;
+
+        let proxy_account_class_hash = get_env_var("PROXY_ACCOUNT_CLASS_HASH")?;
+        let proxy_account_class_hash = FieldElement::from_hex_be(&proxy_account_class_hash).map_err(|_| {
+            ConfigError::EnvironmentVariableSetWrong(format!(
+                "PROXY_ACCOUNT_CLASS_HASH should be provided as a hex string, got {proxy_account_class_hash}"
+            ))
+        })?;
+
+        Ok(StarknetConfig::new(&starknet_rpc_url, kakarot_address, proxy_account_class_hash))
+    }
+}
--- a/crates/rpc/Cargo.toml
+++ b/crates/rpc/Cargo.toml
@@ -35,6 +35,7 @@ starknet = { workspace = true }
 thiserror = "1.0.38"
 tracing = "0.1.34"
 tracing-subscriber = { version = "0.3.3", features = ["env-filter"] }
+lazy_static = { workspace = true }
 
 [dev-dependencies.cargo-husky]
 version = "1.5.0"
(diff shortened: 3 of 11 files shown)


Cargo.toml
```diff
@@ -27,3 +27,4 @@ reth-rlp = { git = "https://github.com/paradigmxyz/reth.git", rev = "fb710e5" }
 reth-primitives = { git = "https://github.com/paradigmxyz/reth.git", rev = "fb710e5" }
 async-trait = "0.1.58"
 jsonrpsee = { version = "0.18.2", features = ["full"] }
+lazy_static = "1.4.0"
```
Eikix, line 30: nice refactor, good thinking

bajpai244: Thanks :)

crates/rpc/src/lib.rs
```diff
@@ -22,7 +24,7 @@ pub enum RpcError {
 ///
 /// Will return `Err` if an error occurs when running the `ServerBuilder` start fails.
 pub async fn run_server(starknet_client: Box<dyn KakarotClient>) -> Result<(SocketAddr, ServerHandle), RpcError> {
-    let socket_addr = std::env::var("KAKAROT_HTTP_RPC_ADDRESS").unwrap_or_else(|_| "0.0.0.0:3030".to_owned());
+    let socket_addr = ENV_KAKAROT_HTTP_RPC_ADDRESS.clone();
```
Eikix, line 27: since this is used in run_server and not main, I think it should be separate

Eikix (changes requested):
As discussed in Daily, let's try to change the approach!:)

crates/core/src/client/config.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub kakarot_address: FieldElement,
+    pub proxy_account_class_hash: FieldElement,
+}
+
+impl StarknetConfig {
+    pub fn new(
+        starknet_rpc: &str,
+        kakarot_address: FieldElement,
+        proxy_account_class_hash: FieldElement,
+    ) -> StarknetConfig {
```
Eikix, line 19: this can be `self`
funny that clippy didnt catch it

bajpai244: Oh! Makes sense!

Yeah, something that should have been caught by Clippy!

crates/core/src/models/cfg.rs
```diff
@@ -0,0 +1,46 @@
+use std::panic;
```
Eikix, line 1: either this or `config.rs` are duplicated, you should probably choose one
additionally, we should try not to use shortened names cfg -> config and StarknetCfg -> StarknetConfig

crates/core/src/models/cfg.rs
```diff
@@ -0,0 +1,46 @@
...
+    Ok(StarknetCfg {
+            starknet_rpc: rpc_url,
+            kakarot_address,
+            proxy_account_class_hash
+        })
+    }
+}
+
+
+// validate_env makes sure that are all required environment variables are set and have valid values.
```
Eikix, line 40: prefer rustdoc style comments

crates/core/src/client/config.rs
```diff
@@ -0,0 +1,37 @@
+use eyre::{eyre, Result};
+use starknet::core::types::FieldElement;
+
+fn get_env_var(name: &str) -> Result<String> {
```
Eikix, line 4: Based on @greged93 's former PRs we could use `thiserror` here to define a configerror, and then add it in the more general error enum as a transparent error

Eikix (changes requested):
this is going in the right direction! ggs

crates/core/src/models/cfg.rs
bajpai244: removed it.

crates/rpc/Cargo.toml
```diff
@@ -36,6 +36,7 @@ starknet = { workspace = true }
 thiserror = "1.0.38"
 tracing = "0.1.34"
 tracing-subscriber = { version = "0.3.3", features = ["env-filter"] }
+lazy_static = { workspace = true }
```
Eikix, line 39: do we still need that?

Eikix:
Careful, big rebase;p

bajpai244:
> Careful, big rebase;p

XD

crates/rpc/Cargo.toml
bajpai244: We can move it back from a workspace dependency to a crate specific dependency.

Eikix (dismissed):
this is great! well done! clean and looking great! 


side quest: optionally, before merge, we could rename all `*_cfg` to `*_config` as I usually prefer non-abbreviated names for variables (e.g. `starknet_config`, `rpc_config`)
to be fair, this is not very impactful, i'm aware that rust uses cfg(test) and many other abbreviations so it's fine,

it's simply that in the kakarot-rpc codebase, i notice we either use full names or initials (e.g. rpc) but not abbreviations

bajpai244:
> this is great! well done! clean and looking great!
> 
> side quest: optionally, before merge, we could rename all `*_cfg` to `*_config` as I usually prefer non-abbreviated names for variables (e.g. `starknet_config`, `rpc_config`) to be fair, this is not very impactful, i'm aware that rust uses cfg(test) and many other abbreviations so it's fine,
> 
> it's simply that in the kakarot-rpc codebase, i notice we either use full names or initials (e.g. rpc) but not abbreviations

It makes sense and we should be consistent :)

Changed and pushed!

Eikix (approved):
great job

Issue #200, opened 2023-06-26
feat: introduce mocked jsonrpcclient for unit tests in core crate


Eikix:
as per @danilowhk 's earlier commits, we can use mockall lib or a similar one to have a mocked instance of starknet_client inside kakarot_client

Eikix:
update! we will use Dojo's mockjsonrpc client and try to implement for simple unit tests

PR #214, opened 2023-06-27
feat: create mock json rpc client

Eikix:
Time spent on this PR:

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #200 

# What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

# Other information


crates/core/Cargo.toml
```diff
@@ -19,6 +19,7 @@ reth-rlp = { workspace = true }
 reth-rpc-api = { workspace = true }
 reth-rpc-types = { workspace = true }
 starknet = { workspace = true }
+dojo-test-utils = { workspace = true }
```
jobez, line 22: i think it makes sense to have this as a dev dependency, might save us some headache when we get to the point where we need to build releases

jobez (changes requested):
only note/question is treating the test utils as a dev dependency

crates/core/Cargo.toml
jobez: more clarity around: my concern/question about having a dev-dependency:  

can it give us a clean boundary for test-compilation and release-compilation? 

from what ive been experiencing in rust you can have one entry into cargo work ok, but running cargo in a subcrate will not be able to compile, or developing for release will have a different feature flag which also breaks compilation

PR #218, opened 2023-06-28
refactor: refactor KakarotProvider trait for seperation of responsibilites

bajpai244:
Refactor `KakarotProvider` trait for seperation of responsibilites

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Currently, we have a single `KakarotProvider` trait, which defines methods, that are a combination of Eth methods and Starknet methods, which removes the lines of separation between these individual responsibilities.

Resolves: #198 

# What is the new behavior?

- `KakarotProvider` is now a combination of `KakarotEthApi` and `KakarotStarknetUtils`.
- `KakarotEthApi`: All Ethereum-related concerns live here.
- `KakarotStarknetUtils`: All Starknet concerns live here.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/core/src/client/client_api.rs
+++ b/crates/core/src/client/client_api.rs
@@ -16,20 +16,10 @@ use crate::models::balance::TokenBalances;
 use crate::models::transaction::StarknetTransactions;
 
 #[async_trait]
-pub trait KakarotProvider: Send + Sync {
-    fn kakarot_address(&self) -> FieldElement;
-    fn proxy_account_class_hash(&self) -> FieldElement;
-    fn starknet_provider(&self) -> &JsonRpcClient<HttpTransport>;
-
+pub trait KakarotEthApi: Send + Sync + KakarotStarknetUtils {
     async fn block_number(&self) -> Result<U64, EthApiError>;
 
     async fn transaction_by_hash(&self, hash: H256) -> Result<EtherTransaction, EthApiError>;
-
-    async fn get_eth_block_from_starknet_block(
-        &self,
-        block_id: StarknetBlockId,
-        hydrated_tx: bool,
-    ) -> Result<RichBlock, EthApiError>;
 
     async fn get_code(
         &self,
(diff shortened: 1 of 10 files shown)


crates/core/src/client/client_api.rs
```diff
@@ -16,21 +16,11 @@ use crate::models::balance::TokenBalances;
 use crate::models::transaction::StarknetTransactions;
 
 #[async_trait]
-pub trait KakarotProvider: Send + Sync {
-    fn kakarot_address(&self) -> FieldElement;
-    fn proxy_account_class_hash(&self) -> FieldElement;
-    fn starknet_provider(&self) -> &JsonRpcClient<HttpTransport>;
-
+pub trait KakarotEthApi: Send + Sync + KakarotStarknetUtils {
```
Eikix, line 19: Do we need `Send + Sync` since KakarotStarknetUtils already has them?

crates/eth-rpc/src/lib.rs
```diff
@@ -24,7 +24,7 @@ pub enum RpcError {
 ///
 /// Will return `Err` if an error occurs when running the `ServerBuilder` start fails.
 pub async fn run_server(
-    starknet_client: Box<dyn KakarotProvider>,
```
Eikix, line 27: oops, typo here from former pr, it should be `starknet_client`  -> `kakarot_client`, right?

Eikix (commented):
Great refactor!
Currently, some logic in `eth-rpc` crate is about converting some eth types into starknet types, 
Can you create an issue to refactor that logic into KakarotEthApi? That way, all functions signature will be Eth-compatible in KakarotEthApi.

All the conversion logic should be inside KakarotStarknetUtils trait.

Example:
```rust
// eth-rpc/src/eth_rpc.rs - line 134 - 
   async fn balance(&self, address: Address, block_number: Option<BlockId>) -> Result<U256> {
        let starknet_block_id =
            ethers_block_id_to_starknet_block_id(block_number.unwrap_or(BlockId::Number(BlockNumberOrTag::Latest)))?;

        let balance = self.kakarot_client.balance(address, starknet_block_id).await?;
        Ok(balance)
    }
```

We'd want to directly call kakarot_client.balance with the incoming types, and do the conversion with KakarotStarknetUtils imo

Wdyt

crates/eth-rpc/src/lib.rs
Eikix: i can do it

Eikix (approved):
lgtm

Issue #215, opened 2023-06-28
feat: unify event handling in `transaction_receipt` logic


Eikix:
# Feature Request

**Describe the Feature Request**

Currently, in `KakarotProvider::transaction_receipt`, we handle logs in transaction receipt like

```rust
// Handle events -- Will error if the event is not a Kakarot event
                    let mut logs = Vec::new();

                    // Cannot use `map` because of the `await` call.
                    for event in events {
                        let contract_address = self.safe_get_evm_address(&event.from_address, &starknet_block_id).await;

                        // event "keys" in cairo are event "topics" in solidity
                        // they're returned as list where consecutive values are
                        // low, high, low, high, etc. of the Uint256 Cairo representation
                        // of the bytes32 topics. This recomputes the original topic
                        let topics = (0..event.keys.len())
                            .step_by(2)
                            .map(|i| {
                                let next_key = *event.keys.get(i + 1).unwrap_or(&FieldElement::ZERO);

                                // Can unwrap here as we know 2^128 is a valid FieldElement
                                let two_pow_16: FieldElement =
                                    FieldElement::from_hex_be("0x100000000000000000000000000000000").unwrap();

                                // TODO: May wrap around prime field - Investigate edge cases
                                let felt_shifted_next_key = next_key * two_pow_16;
                                event.keys[i] + felt_shifted_next_key
                            })
                            .map(|topic| H256::from(&topic.to_bytes_be()))
                            .collect::<Vec<_>>();

                        let data = vec_felt_to_bytes(event.data);

                        let log = Log {
                            // TODO: fetch correct address from Kakarot.
                            // Contract Address is the account contract's address (EOA or KakarotAA)
                            address: H160::from_slice(&contract_address.0),
                            topics,
                            data: RpcBytes::from(data.0),
                            block_hash: None,
                            block_number: None,
                            transaction_hash: None,
                            transaction_index: None,
                            log_index: None,
                            removed: false,
                        };

                        logs.push(log);
                    }
```

We need to refactor this to use @jobez 's code in #174

bajpai244:
So, if I get it right, the refactor is about using `to_eth_log` in the loop to convert events to logs, rather than creating them the way we are doing here, am I right here?

<img width="1426" alt="image" src="https://github.com/kkrt-labs/kakarot-rpc/assets/41180869/d34fbda6-4e46-4ed5-b823-a429f97f9394">

Eikix:
yes exactly

[...]

PR #221, opened 2023-06-28
refactor: use starknet event wrapper in `transaction_receipt`

bajpai244:
Use starknet event wrapper in `transaction_receipt`

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

We are not leveraging the Starknet Wrapper and the `to_eth_log` method implemented in #174 
 
Resolves: #215 

# What is the new behavior?

- We use the wrapper type for native Starknet event
- We use the `to_eth_log` method
- We filter out non-kakarot events from the log
- ~~Updated mock response file to use correct Kakarot_address, so that the tests don't panic with errors defined [here](https://github.com/kkrt-labs/kakarot-rpc/blob/fd9fbf24da43fdd3dd5fff059dabe0b204758b59/crates/core/src/models/event.rs#L34).~~ { as discovered after review from @jobez , that the file can be left as it is, and will act as a test that we are filtering out Starknet events in the RPC call. }

- [ ] Yes
- [x] No

### DIFF
--- a/crates/core/src/client/client_api.rs
+++ b/crates/core/src/client/client_api.rs
@@ -16,7 +16,7 @@ use crate::models::balance::TokenBalances;
 use crate::models::transaction::StarknetTransactions;
 
 #[async_trait]
-pub trait KakarotEthApi: KakarotStarknetUtils {
+pub trait KakarotEthApi: KakarotStarknetApi {
     async fn block_number(&self) -> Result<U64, EthApiError>;
 
     async fn transaction_by_hash(&self, hash: H256) -> Result<EtherTransaction, EthApiError>;
@@ -81,7 +81,7 @@ pub trait KakarotEthApi: KakarotStarknetUtils {
 }
 
 #[async_trait]
-pub trait KakarotStarknetUtils: Send + Sync {
+pub trait KakarotStarknetApi: Send + Sync {
     fn kakarot_address(&self) -> FieldElement;
     fn proxy_account_class_hash(&self) -> FieldElement;
     fn starknet_provider(&self) -> &JsonRpcClient<HttpTransport>;
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -434,11 +434,11 @@ impl KakarotEthApi for KakarotClient<JsonRpcClient<HttpTransport>> {
                     // Cannot use `map` because of the `await` call.
                     for event in events {
                         let event = StarknetEvent::new(event);
-                        let log = event.to_eth_log(self, block_hash, block_number, transaction_hash, None, None).await;
-
-                        if let Ok(log) = log {
+                        if let Ok(log) =
+                            event.to_eth_log(self, block_hash, block_number, transaction_hash, None, None).await
+                        {
                             logs.push(log);
-                        }
+                        };
                     }
 
                     TransactionReceipt {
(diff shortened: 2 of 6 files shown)


bajpai244:
Also, the below error message has a grammatical mistake, it should "a kakarot", and not "an kakarot".

https://github.com/kkrt-labs/kakarot-rpc/blob/fd9fbf24da43fdd3dd5fff059dabe0b204758b59/crates/core/src/models/event.rs#L40

Fixing this as well.

jobez (changes requested):

crates/core/src/client/mod.rs
```diff
@@ -483,43 +483,9 @@ impl KakarotProvider for KakarotClient<JsonRpcClient<HttpTransport>> {
...
-                            block_hash: None,
-                            block_number: None,
-                            transaction_hash: None,
-                            transaction_index: None,
-                            log_index: None,
-                            removed: false,
-                        };
+                        let event = StarknetEvent::new(event);
+                        let log =
+                            event.to_eth_log(self, block_hash, block_number, transaction_hash, None, None).await?;
```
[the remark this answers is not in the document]
bajpai244: Yes, it makes sense, I have refactored.

- So, I also noticed that the mock response for the RPC call doesn't need to be changed! Rather than throwing for this event, it should get filtered out { initially I thought that was a mistake }. I have reverted that file back to the old state.
- As we discussed that in this case any error from `to_eth_log` represents the fact that the log needs to be filtered out, so I have refactored the code accordingly.

crates/core/src/client/mod.rs
```diff
@@ -483,45 +483,12 @@ impl KakarotProvider for KakarotClient<JsonRpcClient<HttpTransport>> {
...
-                            transaction_index: None,
-                            log_index: None,
-                            removed: false,
-                        };
-
-                        logs.push(log);
+                        let event = StarknetEvent::new(event);
+                        let log = event.to_eth_log(self, block_hash, block_number, transaction_hash, None, None).await;
+
+                        if let Ok(log) = log {
```
Eikix, line 489: can you directly write 

`if let Ok(log) = event.to_eth_log(self, block_hash, block_number, transaction_hash, None, None).await `

or compiler is unhappy?

bajpai244: It shows a syntax error

bajpai244: Oh! I made a mistake!

bajpai244: still not satisfied!
Screenshot 2023-06-29 at 7 09 08 PM

crates/core/src/client/client_api.rs
```diff
@@ -16,7 +16,7 @@ use crate::models::balance::TokenBalances;
 use crate::models::transaction::StarknetTransactions;
 
 #[async_trait]
-pub trait KakarotEthApi: KakarotStarknetUtils {
+pub trait KakarotEthApi: KakarotStarknetApi {
```
bajpai244, line 19: Renamed from KakarotStarknetUtils -> KakarotStarknetApi

Eikix (approved):
great job!

jobez (approved):
i approve this pr

PR #222, opened 2023-06-28
feat: try_into for felt -> address

jobez:
# Pull Request type

- [ ] Bugfix
- [ ] Feature
- ✅  Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Testing

Resolves: #213 

# What is the new behavior?

going from felt -> ethereum address is implemented via a `try_into` that asserts that the felt doesn't exceed the expected size.

### DIFF
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -17,7 +17,7 @@ use helpers::{
 // TODO: all reth_primitives::rpc types should be replaced when native reth Log is implemented
 // https://github.com/paradigmxyz/reth/issues/1396#issuecomment-1440890689
 use reth_primitives::{
-    keccak256, Address, BlockId, BlockNumberOrTag, Bloom, Bytes, TransactionSigned, H160, H256, U128, U256, U64, U8,
+    keccak256, Address, BlockId, BlockNumberOrTag, Bloom, Bytes, TransactionSigned, H256, U128, U256, U64, U8,
 };
 use reth_rlp::Decodable;
 use reth_rpc_types::{
@@ -417,7 +417,7 @@ impl KakarotEthApi for KakarotClient<JsonRpcClient<HttpTransport>> {
                             )))?;
 
                             let evm_address = Felt252Wrapper::from(*evm_address);
-                            Some(H160::from(evm_address))
+                            Some(evm_address.try_into()?)
                         }
                     };
 
--- a/crates/core/src/models/felt.rs
+++ b/crates/core/src/models/felt.rs
@@ -1,4 +1,4 @@
-use reth_primitives::{H160, H256, U256};
+use reth_primitives::{Address, H256, U256};
 use starknet::core::types::{FieldElement, FromByteArrayError};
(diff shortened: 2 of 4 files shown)


crates/core/src/client/mod.rs
```diff
@@ -417,7 +417,7 @@ impl KakarotEthApi for KakarotClient<JsonRpcClient<HttpTransport>> {
                             )))?;
 
                             let evm_address = Felt252Wrapper::from(*evm_address);
-                            Some(H160::from(evm_address))
+                            Some(evm_address.try_into()?)
```
jobez, line 420: removed h160 because its the implementation detail, Address is the most meaningful type

crates/core/src/models/felt.rs
```diff
@@ -8,6 +8,11 @@ use crate::client::errors::EthApiError;
 pub enum Felt252WrapperError {
     #[error(transparent)]
     FromByteArrayError(#[from] FromByteArrayError),
+    #[error(
+        "Failed to convert Felt252Wrapper to Ethereum address: the value exceeds the maximum size of an Ethereum \
```
jobez, line 12: error message is in terms of Address

crates/core/src/models/felt.rs
```diff
@@ -38,17 +43,26 @@ impl From<u64> for Felt252Wrapper {
...
+    fn from(address: Address) -> Self {
+        let felt = FieldElement::from_byte_slice_be(&address.0).unwrap(); // safe unwrap since H160 is 20 bytes
         Self(felt)
     }
 }
 
-impl From<Felt252Wrapper> for H160 {
-    fn from(felt: Felt252Wrapper) -> Self {
+impl TryFrom<Felt252Wrapper> for Address {
+    type Error = Felt252WrapperError;
```
jobez, line 54: we want conversion to fail when we are going from felt252wrapper to address assuming

- the felt comes from our kkrt evm in the places where it is expected to properly go from hash -> address
  - https://github.com/kkrt-labs/kakarot/blob/1a52997544ed0f00751e5ba10970d7a8c2c51b32/src/utils/utils.cairo#L792-L799

so if we are getting a felt that is larger than the expected amount, we should throw an error to signal that something is off sooner rather than later

jobez (commented):
the idea here is 

- in places where we are going from felt -> address, the kkrt evm makes sure to do the conversion 
  - https://github.com/kkrt-labs/kakarot/blob/1a52997544ed0f00751e5ba10970d7a8c2c51b32/src/utils/utils.cairo#L792-L799

so if we are getting a felt that is larger than the expected amount, we should throw an error to signal that something is off sooner rather than later

crates/core/src/client/mod.rs
Eikix: yes for sure

crates/core/src/models/felt.rs
Eikix: yes

Eikix (approved):
lgtm

Issue #207, opened 2023-06-26
feat: establish using katana TestSequencer in core crate


jobez:
# Feature Request

**Describe the Feature Request**

We should use test sequencer in core crate for now

not in rpc crate yet (in a second phase yes)

So we should use test sequencer in order to spawn a kakarot client,
have a sequence of actions done using starknet-rs, like deploy kakarot, deploy a few contracts, do something, and then call a method of kakarotclient and expect the correct outcome

one example that comes to mind is for instance, 
deploy kakarot,
deploy counter,
call counter.count(), expect 0
sendrawtx to incr(),
call count.count(), expect 1

**Additional Context**

- need to research the cleanest way to have the kakarot evm's contract artifacts available for this process
  - katana having persistent state is a way to have it done in the kkrt repo via a ci job: https://github.com/dojoengine/dojo/issues/370

PR #223, opened 2023-06-29
establish usage of testsequencer (#207)

jobez:
Time spent on this PR:

# Pull Request type

- [ ] Bugfix
- ✅  Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #207 

# What is the new behavior?

-
-
-

- [ ] Yes
- ✅  No

# Other information
will abstract the steps necessary to be able to run the test in a make target or whatever the rust equivalent is. documenting them here if anyone wants to test locally 

- in order to run crates/core/tests/client.rs::test_counter
  - remove the #[ignore] above it (temporary until CI solution)
  - define `KAKAROT_EVM_PROJECT_DIR` in a `.env` file
  - copy the defined `COMPILED_KAKAROT_PATH` assignment in `.env.example` to your .env
  - build the solidity file
    - run `forge build --names --force`
      - if forge is not installed:
        - ` curl -L https://foundry.paradigm.xyz | bash `
        - foundry up
  - run `./scripts/copy_kakarot_deploy_artifacts`
  - run `cargo test test_counter`


crates/eth-rpc/tests/rpc.rs
```diff
@@ -1,20 +1,93 @@
...
     use starknet::macros::felt;
 
+    use crate::kkrt_prep::{create_call_request, create_raw_tx, init_kkrt_state};
     use crate::utils::setup_kakarot_eth_rpc;
 
+    #[tokio::test]
+    #[ignore]
+    async fn use_test_sequencer_test() {
+        let sn_test_sequencer = TestSequencer::start().await;
+        // TODO: lift from env
```
jobez, line 29: having an env file for these tests that has 

- the evm private key 
- local kakarot repo 
  - can take an optional commit value 
  - script can build kkrt at a particular commit and move it to where the test system needs it

crates/eth-rpc/tests/rpc.rs
```diff
@@ -1,20 +1,93 @@
...
+        let kkrt_eoa_private = H256::from_slice(&hex::decode("bar").unwrap());
+
+        let expected_funded_amount = FieldElement::from_dec_str("100000000000000000").unwrap();
+
+        let (kkrt_address, proxy_account_class_hash, _sn_eoa_address, deployed_addresses) = init_kkrt_state(
+            &sn_test_sequencer,
+            kkrt_eoa_address,
+            kkrt_eoa_private,
+            expected_funded_amount,
+            "Counter.json",
```
jobez, line 40: can look at the kakarot repo for example structure where we can commit solidity files here

crates/eth-rpc/tests/rpc.rs
```diff
@@ -1,20 +1,93 @@
...
+        let counter_eth_addr = {
+            // TODO  use try_into here when merged
+            let address = &deployed_addresses.first().unwrap().to_bytes_be()[12..];
+            Address::from_slice(address)
+        };
+
+        // asserts there is code at this address
+        // let code_res = kkrt_eth_rpc.get_code(counter_eth_addr, Option::None).await;
+
+        let inc_selector = hex!("371303c0");
```
jobez, line 74: i sourced the selectors from the compiled json, need to look for the equivalent `get_selector_from_name` helper in reth

crates/eth-rpc/tests/rpc.rs
```diff
@@ -1,20 +1,93 @@
...
+        let inc_tx = create_raw_tx(inc_selector, kkrt_eoa_private, counter_eth_addr, vec![], 1u64);
```
jobez, line 75: manually setting nonce here for the raw transaction, should actually retrieve it

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+        .unwrap();
+
+    // TODO: add check here
+    let _deployment_of_eoa_account_result_receipt =
+        account.provider().get_transaction_receipt(deployment_of_eoa_account_result.transaction_hash).await.unwrap();
+
+    let eoa_account_starknet_address = *eoa_account_starknet_address_result.unwrap().first().unwrap();
+
+    let amount_high = FieldElement::from_dec_str("0").unwrap();
+    let transfer_calldata = vec![eoa_account_starknet_address, amount, amount_high];
```
jobez, line 276: should convey the transfer amount to the deployed evm account as a uint256

jobez (commented):
- next iteration
  - ✅  clean up todos
  - add function comments. i left one as an example for review to see if that was the right format
  - ✅  add the ability to pass in constructor arguments for eth deploys
  - ✅  impl the equivalent of an env file or constants file for 
    - the eth account(s) we need to deploy
- ✅  script that can help copy build artifacts from kakarot-evm to the cli locally
    -  need to look into github actions   for the ci
- ✅ similar structure to kakarot/solidity_contracts 
    - need to look into github actions for the ci 

 

lmk if you see anything else as a blocker

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+use starknet::providers::{JsonRpcClient, Provider};
+use starknet::signers::{LocalWallet, SigningKey};
+use url::Url;
+
+pub fn get_contract(filename: &str) -> (Abi, EthersBytes) {
+    let path = format!("./tests/compiled_solidity/{filename}");
+    let contents = fs::read_to_string(path).unwrap();
+    let obj: JsonAbi = serde_json::from_str(&contents).unwrap();
+    let JsonAbi::Object(obj) = obj else { panic!() };
+    (serde_json::from_str(&serde_json::to_string(&obj.abi).unwrap()).unwrap(), obj.bytecode.unwrap().to_vec())
```
Eikix, line 31: why is this serialization to deserialization (to switch types) necessary? I understand we have a obj of type `JsonAbi::Object(obj)`

Can't we access the `Abi` type from a JsonAbi::Object?

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+}
+
+pub fn create_call_request(to: Address, selector: [u8; 4]) -> CallRequest {
+    CallRequest {
+        from: Option::None,
```
Eikix, line 36: do we need to specify Option::None (versus None)

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
+use std::collections::HashMap;
```
Eikix, line 1: this file should probably not be called kkrt_prep 🥕
inspiration:
script_utils 
utils
helpers
deploy_helpers
tx_helpers

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+    let mut raw_tx = BytesMut::new(); // Create a new empty buffer
+
+    signed_transaction.encode_enveloped(&mut raw_tx); // Encode the transaction into the buffer
+
+    raw_tx.to_vec().into()
+}
+
+async fn deploy_kkrt_eth_contract(
+    sequencer_url: Url,
+    deployed_eao_starknet_account: FieldElement,
```
Eikix, line 96: imo: `deployed_eao_starknet_account` -> `eoa_account_starknet_address`

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+    eoa_secret: H256,
```
Eikix, line 97: `eoa_ethereum_private_key`

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    let field_elements: Vec<FieldElement> = bytes_vec
+        .into_iter()
+        .map(|byte| {
+            // Convert each byte to a field element
+            FieldElement::from(byte)
+        })
+        .collect();
+
+    let not_used_eoa_account_fields = FieldElement::from_hex_be("0xDEAD").unwrap();
```
Eikix, line 156: can be called `unused_eoa_field` -> FieldElement::zero

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+    contract_name: &str,
+) -> Option<Vec<FieldElement>> {
+    // need contract artifact
+    // turns into the data field of a tx_payload
+    // that gets signed and executed in the starknet eoa 'execute'/invoke
+    // can first experiment with doing a call for `get_nonce`
+
+    let signing_key = SigningKey::from_secret_scalar(FieldElement::from_byte_slice_be(eoa_secret.as_slice()).unwrap());
+
+    let eoa_starknet_account = SingleOwnerAccount::new(
```
Eikix, line 107: I'm a bit confused by this -> we can construct a starknet account struct, that has an ECDSA signing key, and perform calls with it?

Seems a bit misleading since the only signature that is "considered" by EOA is the one inside the RLP encoded signed tx (inside calldata), rather than in signed field.

But also, how does `let deployment_of_counter_evm_contract_result = eoa_starknet_account
        .execute(vec![Call {
            calldata: field_elements,
            to: not_used_eoa_account_fields,
            selector: not_used_eoa_account_fields,
        }])` sign the payload? Does it know a signature scheme?

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+        JsonRpcClient::new(HttpTransport::new(sequencer_url)),
```
Eikix, line 108: yayy

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    let signed_transaction = TransactionSigned::from_transaction_and_signature(transaction, signature);
+    let mut raw_tx = BytesMut::new(); // Create a new empty buffer
+
+    signed_transaction.encode_enveloped(&mut raw_tx); // Encode the transaction into the buffer
+
+    raw_tx.to_vec().into()
+}
+
+async fn deploy_kkrt_eth_contract(
```
Eikix, line 94: `deploy_kkrt_eth_contract` -> `deploy_evm_contract`

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+    let contract_address =
+        get_contract_address(FieldElement::ZERO, *class_hash, &constructor_calldata.clone(), FieldElement::ZERO);
+
+    // TODO: add a check here
+    let _receipt = account.provider().get_transaction_receipt(result.transaction_hash).await.unwrap();
+
+    Ok(contract_address)
+}
+
+async fn declare_kkrt_contracts(
```
Eikix, line 208: kkrt -> kakarot
in general, let's prefer full words rather than initials

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+    account: &SingleOwnerAccount<JsonRpcClient<HttpTransport>, LocalWallet>,
+) -> HashMap<String, FieldElement> {
+    let paths = fs::read_dir("tests/compiled_kkrt").expect("Could not read directory");
```
Eikix, line 211: `tests/compiled_kakarot`

i'd suggest putting the paths in module constants (with const keyword, &str, and fullcaps)

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    let kkrt_compiled_contracts: Vec<_> = paths
```
Eikix, line 213: Interesting!

Eikix: let's rename `kakarot_compiled_contract_files` or `kakarot_compiled_contract_paths`

crates/eth-rpc/tests/kkrt_prep.rs
jobez: i yoinked this as an example from ethers, will revisit to see if theres something cleaner

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+        entry_point_selector: get_selector_from_name("balanceOf").unwrap(),
+        calldata: vec![eoa_account_starknet_address],
+    };
+
+    let _balance_of_starknet_address_of_eoa_result =
+        account.provider().call(call_get_balance_of_starknet_address_of_eoa, BlockId::Tag(BlockTag::Latest)).await;
+
+    let call_get_nonce_of_starknet_address_of_eoa = FunctionCall {
+        contract_address: eoa_account_starknet_address,
+        entry_point_selector: get_selector_from_name("get_nonce").unwrap(),
```
Eikix, line 303: what is the difference between selector! and get_selector_from_name

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+    let call_get_balance_of_starknet_address_of_eoa = FunctionCall {
+        contract_address: fee_token_address,
+        entry_point_selector: get_selector_from_name("balanceOf").unwrap(),
+        calldata: vec![eoa_account_starknet_address],
+    };
+
+    let _balance_of_starknet_address_of_eoa_result =
+        account.provider().call(call_get_balance_of_starknet_address_of_eoa, BlockId::Tag(BlockTag::Latest)).await;
+
+    let call_get_nonce_of_starknet_address_of_eoa = FunctionCall {
```
Eikix, line 301: I think we don't need to do this, we can simply get the nonce from the starknet client itself -> we should deprecate this nonce function soon

crates/eth-rpc/tests/kkrt_prep.rs
```diff
@@ -0,0 +1,377 @@
...
+        chain_id::TESTNET,
+    );
+
+    // TODO: dehardcode
+    let chain_id = 1263227476;
+    let nonce = 0;
+    let gas = 1000;
+
+    let (_abi, contract_bytes) = get_contract(contract_name);
+    let contract_bytes_reth: reth_primitives::Bytes = contract_bytes.into(); // convert Vec<u8> to reth_primitives::Bytes
```
jobez, line 120: make it so contract bytes also has constructor that is encoded

crates/eth-rpc/tests/kkrt_prep.rs
jobez: per our short call, will update this so its more clear and point to where people can get a clear view of how we handle eth payloads -> starknet transactions

```
    // This a made up signing key so we can reuse starknet-rs abstractions
    // to see the flow of how kakarot-rpc handles eth payloads -> starknet txns
    // see ./crates/core/src/client.rs::send_transaction 
    let signing_key = SigningKey::from_secret_scalar(FieldElement::ZERO);

    let eoa_starknet_account = SingleOwnerAccount::new(
        JsonRpcClient::new(HttpTransport::new(sequencer_url)),
        LocalWallet::from_signing_key(signing_key),
        eoa_account_starknet_address,
        chain_id::TESTNET,
    );

```

crates/eth-rpc/tests/kkrt_prep.rs
jobez: https://github.com/xJonathanLEI/starknet-rs/blob/master/starknet-macros/src/lib.rs#L8-L27

selector! is a process macro that internally uses get_selector_from_name

.env.example
```diff
@@ -4,3 +4,7 @@ RUST_LOG=debug
 KAKAROT_HTTP_RPC_ADDRESS=0.0.0.0:3030
 KAKAROT_ADDRESS=0x7a88f6f9d63ccaa5855babb32cbb0230b8588aaaa6bc4ce2d173fa528ce7567
 PROXY_ACCOUNT_CLASS_HASH=0x3010a53967fa04842bcbcb6de8817101f047ef0d074b3eacbe714a3fc42a2eb
+
+# configurations for testing
+COMPILED_KAKAROT_PATH=crates/core/tests/compiled_solidity
+KAKAROT_EVM_PROJECT_DIR=
```
Eikix, line 10: `KAKAROT_EVM_PROJECT_DIR` what does this mean?
is this simply the folder where smart contracts live?

Could just call it `SOLIDITY_CONTRACTS` then to match the folder you committed

Cargo.toml
```diff
@@ -29,6 +29,7 @@ reth-primitives = { git = "https://github.com/paradigmxyz/reth.git", rev = "fb71
 async-trait = "0.1.58"
 jsonrpsee = { version = "0.18.2", features = ["full"] }
 lazy_static = "1.4.0"
+dotenv = "0.15.0"
```
Eikix, line 32: nice:)

crates/core/tests/helpers/constants.rs
```diff
@@ -0,0 +1,4 @@
+pub const EVM_ADDRESS: &str = "0x132d200e5ceedf1a4634eee081cfeb077d92e4fd";
```
Eikix, line 1: I think we can use lazy_static! here to make those constants directly the desired type, we've done it once already in the codebase

crates/core/tests/client.rs
```diff
@@ -1,22 +1,137 @@
...
     use reth_rpc_types::Log;
     use starknet::core::types::{BlockId, BlockTag, Event, FieldElement};
     use starknet::core::utils::get_selector_from_name;
     use starknet::providers::Provider;
 
+    use crate::helpers::constants::{EVM_ADDRESS, EVM_PRIVATE_KEY};
+    use crate::helpers::deploy_helpers::{create_raw_ethereum_tx, deploy_kakarot_system};
+
+    #[tokio::test]
+    #[ignore]
```
Eikix, line 30: why are we ignoring this test?

crates/core/tests/client.rs
```diff
@@ -1,22 +1,137 @@
...
+        let deployed_kakarot =
+            deploy_kakarot_system(&starknet_test_sequencer, EVM_ADDRESS, eoa_private, expected_funded_amount).await;
+
+        let address = "0x0000000000000000000000000000000000000000".parse::<EthersAddress>().unwrap();
+        let (_constructable_abi, deployed_addresses) = deployed_kakarot
+            .deploy_evm_contract(
+                starknet_test_sequencer.url(),
+                "Constructable",
+                // more than one argument to a constructor needs to be conveyed as a tuple
+                (EthersU256::from(100), address),
```
Eikix, line 48: Why are we using a ethers-rs type here?

crates/core/tests/client.rs
```diff
@@ -1,22 +1,137 @@
...
+            .unwrap();
+
+        let kakarot_client = KakarotClient::new(StarknetConfig::new(
+            starknet_test_sequencer.url().as_ref(),
+            deployed_kakarot.kakarot,
+            deployed_kakarot.kakarot_proxy,
+        ))
+        .unwrap();
+
+        let Ok(opcode_eth_addr) = ({
```
Eikix, line 60: opcode_eth_addr -> `constructable_contract_address`

crates/core/tests/client.rs
```diff
@@ -1,22 +1,137 @@
...
+            let address : Felt252Wrapper = (*deployed_addresses.first().unwrap()).into();
```
Eikix, line 61: We could refactor this to be: 
```rust
let constructable_contract_address = Felt252Wrapper::try_from((*deployed_addresses.first().unwrap()).into()).unwrap();
```

crates/core/tests/client.rs
```diff
@@ -1,22 +1,137 @@
...
+        kakarot_client.get_code(opcode_eth_addr, BlockId::Tag(BlockTag::Latest)).await.expect("contract not deployed");
+    }
+
+    #[tokio::test]
+    #[ignore]
+    async fn test_counter() {
+        let starknet_test_sequencer = TestSequencer::start().await;
+
+        let eoa_public = Address::from_slice(&hex::decode(EVM_ADDRESS.trim_start_matches("0x")).unwrap());
+        let eoa_private = H256::from_slice(&hex::decode(EVM_PRIVATE_KEY).unwrap());
```
Eikix, line 76: we should strive to make things as clear as possible, even when it's seemingly obvious, because in other places, it may not be obvious for some

eoa_private -> `eoa_private_key`

crates/core/tests/client.rs
```diff
@@ -1,22 +1,137 @@
...
+
+        kakarot_client.get_code(opcode_eth_addr, BlockId::Tag(BlockTag::Latest)).await.expect("contract not deployed");
+    }
+
+    #[tokio::test]
+    #[ignore]
+    async fn test_counter() {
+        let starknet_test_sequencer = TestSequencer::start().await;
+
+        let eoa_public = Address::from_slice(&hex::decode(EVM_ADDRESS.trim_start_matches("0x")).unwrap());
```
Eikix, line 75: `eoa_public` -> `eoa_evm_address` or `eoa_eth_address`

Eikix: after giving it some thought, i think you're right in naming things with eth_ prefix for addresses and so on, evm doesn't make so much sense

so for  every x, eth address can be `x_eth_address` 

here -> `eoa_eth_address`

crates/core/tests/client.rs
```diff
@@ -1,22 +1,137 @@
...
+        let inc_tx =
+            create_raw_ethereum_tx(inc_selector, eoa_private, counter_eth_addr, vec![], nonce.try_into().unwrap());
+        let inc_res = kakarot_client.send_transaction(inc_tx).await.unwrap();
+        kakarot_client.transaction_receipt(inc_res).await.expect("increment receipt failed");
+
+        let count_selector = counter_abi.function("count").unwrap().short_signature();
+        let counter_val =
+            kakarot_client.call_view(counter_eth_addr, count_selector.into(), BlockId::Tag(BlockTag::Latest)).await;
+
+        if let Ok(bytes) = counter_val {
```
Eikix, line 127: You can `counter_val.expect(<error msg goes here>)` instead of the if elses imho

Eikix: even just unwrapping here is fine

crates/core/tests/helpers/deploy_helpers.rs
```diff
@@ -0,0 +1,476 @@
...
+        // this is done so we can have more flexibility in filepaths the artifacts are located
+        // so multiple crates can access them, if necessary
+        let project_root = crate_root.parent().unwrap().parent().unwrap();
+        let full_path = project_root.join($relative_path);
+        full_path
+    }};
+}
+
+#[derive(Debug, Deserialize)]
+struct Profile {
```
Eikix, line 47: could you add some contents using rustdoc comments?

crates/core/tests/helpers/deploy_helpers.rs
```diff
@@ -0,0 +1,476 @@
...
+struct DefaultProfile {
+    out: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct FoundryConfig {
+    profile: Profile,
+}
+
+fn get_foundry_default_out(file_path: &std::path::Path) -> Result<String, Box<dyn std::error::Error>> {
```
Eikix, line 61: could you add some context, using rustdoc comments?

also, I think you could technically hardcode the foundry.toml filename / path, since we're forcing this util to be coupled with foundry types `FoundryConfig`

crates/core/tests/helpers/deploy_helpers.rs
```diff
@@ -0,0 +1,476 @@
...
+    file.read_to_string(&mut contents)?;
+
+    let config: FoundryConfig = toml::from_str(&contents)?;
+
+    // Since 'out' is a field in 'Default' struct, we can access it directly
+    Ok(config.profile.default.out)
+}
+
+// This assumes you are adding a solidity file in kakarot-rpc/solidity_contracts
+// and ran `forge build --names --force`
```
Eikix, line 73: why do we need to make that assumption?

Eikix: I think all utils deserve rustdoc style comments to help give context on what they do/achieve

crates/core/tests/helpers/deploy_helpers.rs
```diff
@@ -0,0 +1,476 @@
...
+        (None, true) => bytecode.clone(),
+        (Some(constructor), _) => {
+            let res = constructor.encode_input(bytecode.to_vec(), &params);
+            res.unwrap().into()
+        }
+    }
+}
+
+// Constructs an ethereum transaction with the correct Kakarot chain ID, and default values
+// for everything but nonce, to, and bytes
```
Eikix, line 120: why is `to` field not an Address?

crates/core/tests/helpers/deploy_helpers.rs
```diff
@@ -0,0 +1,476 @@
...
+        max_fee_per_gas: Default::default(),
+        gas_limit: Default::default(),
+        to,
+        value: Default::default(),
+        input,
+        access_list: Default::default(),
+    })
+}
+
+pub fn create_raw_ethereum_tx(selector: [u8; 4], eoa_secret: H256, to: Address, args: Vec<U256>, nonce: u64) -> Bytes {
```
Eikix, line 135: `eoa_secret` -> `eoa_private_key` to keep consistency

crates/core/tests/helpers/deploy_helpers.rs
Eikix: Ahh I think I understand, because TransactionKind::create will probably be None or smth

crates/core/tests/client.rs
jobez: Those types are the ones that implement the Tokenizable trait, which handles the move from ethers types -> byte representation when we include it in the contract bytecode.

crates/core/tests/helpers/deploy_helpers.rs
jobez: yeah, reth implements this as an enum type: https://github.com/paradigmxyz/reth/blob/main/crates/primitives/src/transaction/mod.rs#L716

crates/core/tests/client.rs
jobez: The tests are ignored because the kakarot artifacts need to be moved to wherever we define `COMPILED_KAKAROT_PATH` in order to successfully deploy. I added the `KAKAROT_EVM_PROJECT_DIR` and the `scripts/copy_kakarot_deploy_artifacts.sh` with the thought that this can help the flow locally and also to do it in a github action before the tests run.

Wdyt? Follow me here? Perhaps there is a simpler option!

crates/core/tests/client.rs
jobez: is it ok if I do this? I think trying to do it all in one line requires adding more type info and I find this easier to understand:

```
        let counter_eth_address = {
            let address : Felt252Wrapper = (*deployed_addresses.first().unwrap()).into();
            address.try_into().unwrap()
        };
```

.env.example
jobez: its used by the `scripts/copy_kakarot_deploy_artifacts.sh` to build kakarot v0 contracts and move its deploy artifacts to where the tests expect them (the `COMPILED_KAKAROT_PATH`)

Is this sensible/clear? 👂

crates/core/tests/client.rs
Eikix: I see!

crates/core/tests/client.rs
```diff
@@ -1,22 +1,137 @@
...
+
+        let eoa_private = H256::from_slice(&hex::decode(EVM_PRIVATE_KEY).unwrap());
+
+        let expected_funded_amount = FieldElement::from_dec_str("100000000000000000").unwrap();
+
+        let deployed_kakarot =
+            deploy_kakarot_system(&starknet_test_sequencer, EVM_ADDRESS, eoa_private, expected_funded_amount).await;
+
+        let address = "0x0000000000000000000000000000000000000000".parse::<EthersAddress>().unwrap();
+        let (_constructable_abi, deployed_addresses) = deployed_kakarot
```
ClementWalter, line 43: why do we deploy using kakarot directly and not using evm things?

crates/core/tests/client.rs
ClementWalter, line 75: can't we avoid all these conversions for going from `EVM_ADDRESS` to `eoa_public`?

crates/core/tests/helpers/constants.rs
ClementWalter: and `EVM_ADDRESS` should be derived from the `EVM_PRIVATE_KEY` to prevent any mismatch

crates/core/tests/helpers/deploy_helpers.rs
```diff
@@ -0,0 +1,476 @@
...
+use starknet::signers::{LocalWallet, SigningKey};
+use toml;
+use url::Url;
+
+use super::constants::FEE_TOKEN_ADDRESS;
+
+macro_rules! root_project_path {
+    ($relative_path:expr) => {{
+        let crate_root = std::path::Path::new(env!("CARGO_MANIFEST_DIR"));
+        // this is assuming this code will always be ran in the context where the CARGO_MANIFEST_DIR
```
ClementWalter, line 36: I don't understand this sentence

crates/core/tests/helpers/deploy_helpers.rs
```diff
@@ -0,0 +1,476 @@
...
+
+#[derive(Debug, Deserialize)]
+struct Profile {
+    default: DefaultProfile,
+}
+
+#[derive(Debug, Deserialize)]
+struct DefaultProfile {
+    out: String,
+}
```
ClementWalter, line 54: I find it easier to read when variables/func are defined when possible in order, ie. here

```
struct DefaultProfile {...}
struct Profile {...}
struct FoundryConfig {...}
```

crates/core/tests/helpers/deploy_helpers.rs
```diff
@@ -0,0 +1,476 @@
...
+    // Since 'out' is a field in 'Default' struct, we can access it directly
+    Ok(config.profile.default.out)
+}
+
+// This assumes you are adding a solidity file in kakarot-rpc/solidity_contracts
+// and ran `forge build --names --force`
+pub fn get_contract(filename: &str) -> (Abi, ethers::types::Bytes) {
+    let dot_sol = format!("{filename}.sol");
+    let dot_json = format!("{filename}.json");
+    let foundry_toml_path = root_project_path!("foundry.toml");
```
ClementWalter, line 77: I'm wondering if we could use the forge create directly instead

ClementWalter (commented):
posting WIP comments as I need to leave

crates/core/tests/client.rs
Eikix: not sure about this question but: 
because we don't have access to an ethereum json rpc spec -> we cannot use ether-rs 

we only have access to the kakarot client, it does not implement json rpc of eth

jobez: yeah, that's the crux of it. we are 'inbetween' starknet and ethereum at this stage. 

we would want to test at the rpc level the rpc's ability to do deployment, i agree.

crates/core/tests/client.rs
```diff
@@ -1,20 +1,145 @@
...
+
+    #[tokio::test]
+    async fn test_constructable() {
+        // initial setup of Constructable to test we can deploy contracts w/ constructor arguments
+        let starknet_test_sequencer = TestSequencer::start().await;
+
+        let expected_funded_amount = FieldElement::from_dec_str("100000000000000000").unwrap();
+
+        let deployed_kakarot =
+            deploy_kakarot_system(&starknet_test_sequencer, EOA_WALLET.clone(), expected_funded_amount).await;
```
jobez, line 38: just want to leave a note about why I used `clone` here, as I am beginning to understand Rust's memory management model

just trying to pass `EOA_WALLET` gives me

```
   --> crates/core/tests/client.rs:38:61

    |

38  |             deploy_kakarot_system(&starknet_test_sequencer, EOA_WALLET, expected_funded_amount).await;

    |             ---------------------                           ^^^^^^^^^^ expected `Wallet<SigningKey<Secp256k1>>`, found `EOA_WALLET`

    |             |

    |             arguments to this function are incorrect

    |

    = note: expected struct `Wallet<ecdsa::signing::SigningKey<Secp256k1>>`

               found struct `EOA_WALLET`

```

so then I would want to derefence `*EOA_WALLET` and see:

```
error[E0507]: cannot move out of dereference of `EOA_WALLET`

  --> crates/core/tests/client.rs:38:61

   |

38 |             deploy_kakarot_system(&starknet_test_sequencer, *EOA_WALLET, expected_funded_amount).await;

   |                                                             ^^^^^^^^^^^ move occurs because value has type `Wallet<ecdsa::signing::SigningKey<Secp256k1>>`, which does not implement the `Copy` trait

```

which is what brought me to cloning.

jobez (commented):
ok, now the tests are running on the CI
- we have `make setup` that 
   - pulls kakarot submodule and builds kakarot
   - builds solidity files
   - builds solidity

- so now the setup on top of the previous is
  - install foundry
  - run `make setup`
  - either setup your own env file then run `make test` or call .`/scripts/make_with_env.sh test`

crates/core/tests/helpers/deploy_helpers.rs
jobez: i changed the logic use foundry-config's logic and updated the docstring, lmk if it is more clear

https://github.com/foundry-rs/foundry/blob/42fb69461386d58d6f656019aca665c3aed682ae/config/src/utils.rs#L29-L39

.github/workflows/test.yml
```diff
@@ -63,10 +69,37 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
+      - run: cd kakarot && poetry config installer.modern-installation false
       - name: Generate code coverage
         run:
-          cargo llvm-cov --all-features --workspace --lcov --output-path
-          lcov.info
+          ./scripts/make_with_env.sh test-coverage
```
Eikix, line 102: could this be refactored with:

`env` yaml key in your job in order to source .env variables without using a .sh script?

Eikix: example from kakarot:

```
jobs:
  deploy:
    runs-on: ubuntu-latest
    env:
      ACCOUNT_ADDRESS: ${{ secrets.ACCOUNT_ADDRESS }}
      PRIVATE_KEY: ${{ secrets.PRIVATE_KEY }}
      DEVNET_ACCOUNT_ADDRESS: ${{ secrets.DEVNET_ACCOUNT_ADDRESS }}
      DEVNET_PRIVATE_KEY: ${{ secrets.DEVNET_PRIVATE_KEY }}
      RPC_KEY: ${{ secrets.RPC_KEY }}
      INFURA_KEY: ${{ secrets.INFURA_KEY }}
    outputs:
      devnet-version: ${{ steps.devnet-version.outputs.DEVNET_VERSION }}
    steps:
      - uses: actions/checkout@v3
// etc.
```

Eikix: PS: does this need "make" before test-coverage, or all make commands can just be accessed like this?

jobez: all commands.

the only env variable we need is for testing, so i dont think it makes sense to establish it as a secret. i can remove the env variable and just hardcode that the tests will expect deploy artifacts in the ./katana/build directory. i think that may be the right way forward.

Eikix: alright works for me

.github/workflows/test.yml
```diff
@@ -63,10 +69,37 @@ jobs:
...
+          installer-parallel: true
+      - name: pull kakarot submodule and setup
+        run:
+          make setup
+      - name: Load cached venv
+        id: cached-poetry-dependencies
+        uses: actions/cache@v3
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
```
ClementWalter, line 98: the venv is loaded after `make setup`?

solidity_contracts/Constructable.sol
ClementWalter, line 1: can't we use the contract and foundry.toml from the kakarot submodule?

jobez: yeah, we can actually, np

jobez: but it might be good to maybe have both, so you can have contracts meant for rpc testing? maybe thats overkill.

Eikix: let's use kakarot's submodule for now that way -> we can't have different foundry settings (one source of truth)

but let's use both for smart contracts -> we can use "overcharge" the set of SC but from kakarot-rpc

.github/workflows/test.yml
Eikix: what env variable is make setup using?

jobez: i do not think i am doing this right, i was emulating the kakarot ci and i think there's output in the runner that it isn't setup correctly. will link to it here when i find it again

Eikix (approved):
good job powering through 🥷

PR #262, opened 2023-07-06
bug-fix/use correct nonce in `deploy_evm_contract`

jobez:
Time spent on this PR: .2

# Pull Request type

- [X] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

was using Default::default() for the nonce value that was being constructed in the ethereum transaction that was being sent in a starknet transaction. this breaks when you try to use the starknet address underpinning the ethereum deploy more than once. 

Resolves: #<Issue number>

# What is the new behavior?

now we retrieve the nonce of the `eoa_starknet_account` per each deploy

-
-
-

- [ ] Yes
- [X] No

# Other information

### DIFF
--- a/crates/core/src/test_utils/deploy_helpers.rs
+++ b/crates/core/src/test_utils/deploy_helpers.rs
@@ -225,8 +225,9 @@ async fn deploy_evm_contract<T: Tokenize>(
 
     let (abi, contract_bytes) = get_contract(contract_name);
     let contract_bytes = encode_contract(&abi, &contract_bytes, constructor_args);
+    let nonce = eoa_starknet_account.get_nonce().await.unwrap();
     let transaction =
-        to_kakarot_transaction(Default::default(), TransactionKind::Create, contract_bytes.to_vec().into());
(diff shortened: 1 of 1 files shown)


crates/core/src/test_utils/deploy_helpers.rs
```diff
@@ -225,8 +225,9 @@ async fn deploy_evm_contract<T: Tokenize>(
 
     let (abi, contract_bytes) = get_contract(contract_name);
     let contract_bytes = encode_contract(&abi, &contract_bytes, constructor_args);
+    let nonce = eoa_starknet_account.get_nonce().await.unwrap();
     let transaction =
-        to_kakarot_transaction(Default::default(), TransactionKind::Create, contract_bytes.to_vec().into());
+        to_kakarot_transaction(nonce.try_into().unwrap(), TransactionKind::Create, contract_bytes.to_vec().into());
```
Eikix, line 230: nice catch

Eikix (approved):

Issue #259, opened 2023-07-06
bug: RPC should not raise when EOA are not deployed


ClementWalter:
# Bug Report

**Kakarot RPC version:**

Current

**Current behavior:**

When a client request "eth_getTransactionCount" for an EAO that is not deployed, the RPC raises "contract not found"

**Expected behavior:**

Returns 0

**Steps to reproduce:**

Run the RPC targeting any chain (madara/testnet)
Make a raw request targeting a non deployed EOA

```
curl http://127.0.0.1:3030 \
  -H 'content-type: application/json' \
  --data-raw '{"method":"eth_getTransactionCount","jsonrpc":"2.0","params":["0x1804c8ab1f12e6bbf3894d4083f33e07309d1f38"],"id":0}' \
  --compressed | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   194  100    79  100   115    154    225 --:--:-- --:--:-- --:--:--   383
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32001,
    "message": "Contract not found"
  },
  "id": 0
}
```

**Related code:**

`transaction_count`

**Other information:**

Generally speaking, we should **not** raise when queried with address of non deployed EOA as this concept does not exist in the EVM. Just returns 0, empty, etc.

PR #264, opened 2023-07-07
map `NoContractError` to zero for `eth_getTransactionCount`

jobez:
Time spent on this PR: .1

# Pull Request type

- [X] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

An EOA address is mapped to a starknet address w/ `compute_starknet_address`, then looked up for a nonce against the provider. When the mapped address doesn't exist, we get a NoContractError.

Resolves: #259 

We match for NoContractError's and map it to a zero, per the evm tooling norms.

-
-
-

- [ ] Yes
- [X] No

# Other information

### DIFF
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -21,12 +21,12 @@ use reth_rpc_types::{
 };
 use starknet::core::types::{
     BlockId as StarknetBlockId, BlockTag, BroadcastedInvokeTransaction, BroadcastedInvokeTransactionV1, FieldElement,
-    FunctionCall, InvokeTransactionReceipt, MaybePendingBlockWithTxs, MaybePendingTransactionReceipt, SyncStatusType,
-    Transaction as TransactionType, TransactionReceipt as StarknetTransactionReceipt,
+    FunctionCall, InvokeTransactionReceipt, MaybePendingBlockWithTxs, MaybePendingTransactionReceipt, StarknetError,
+    SyncStatusType, Transaction as TransactionType, TransactionReceipt as StarknetTransactionReceipt,
     TransactionStatus as StarknetTransactionStatus,
 };
 use starknet::providers::jsonrpc::{JsonRpcClient, JsonRpcTransport};
-use starknet::providers::Provider;
+use starknet::providers::{Provider, ProviderError};
 
 use self::api::{KakarotEthApi, KakarotStarknetApi};
 use self::config::StarknetConfig;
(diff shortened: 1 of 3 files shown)


crates/core/tests/client.rs
```diff
@@ -17,13 +17,40 @@ mod tests {
...
-    use reth_primitives::{Address, BlockId, Bytes, H256};
+    use reth_primitives::{Address, BlockId, BlockNumberOrTag, Bytes, H256, U256};
     use reth_rpc_types::Log;
     use starknet::core::types::{BlockId as StarknetBlockId, BlockTag, Event, FieldElement};
     use starknet::core::utils::get_selector_from_name;
     use starknet::providers::jsonrpc::HttpTransport;
     use starknet::providers::{JsonRpcClient, Provider};
 
+    #[tokio::test]
+    async fn test_rpc_should_not_raise_when_eoa_not_deployed() {
```
jobez, line 28: might be overkill to have this test, but having it locally was super useful

crates/core/src/client/mod.rs
```diff
@@ -372,13 +372,23 @@ impl<T: JsonRpcTransport + Send + Sync> KakarotEthApi<T> for KakarotClient<JsonR
...
-        Ok(nonce.into())
+        self.starknet_provider
+            .get_nonce(starknet_block_id, starknet_address)
+            .await
+            .map(|nonce| {
+                let nonce: Felt252Wrapper = nonce.into();
+                nonce.into()
+            })
+            .or_else(|err| match err {
+                ProviderError::StarknetError(StarknetError::ContractNotFound) => Ok(U256::from(0)),
```
jobez, line 389: would it be useful to have dev level logging that prints 'address not found' specifically? i think it probably would be. 

are we doing logging like this anywhere?

Eikix: Ahh good question, I guess not!
We have not setup a dev logging system yet, apart from the tracing subscriber

crates/core/tests/client.rs
Eikix: i like! a good reminder that eoas in EVM all have nonce 0 by default, and don't throw when requiring the nonce of a non-existing addr

crates/core/src/client/mod.rs
jobez: when trying to deploy kakaswap this was one of the error cases i was first hitting when the rpc would 'throw' on trying to get a nonce. having a clear signal to the brave dev attempting to use it for tooling that 'hey, this address doesnt exist, did you deploy?

ClementWalter: approve that non existing EOAccount is n°1 issue for anyone trying to use kakarot-rpc. In the python sandbox, I added a Starknet account owned by the RPC to deploy if required EOAccount.

We don't need to always deploy (like returning 0 is just the good thing to do for nonce), but trying to send a tx, we should, at least in the beginning.

From a UX point of viw, we should not have EOAccount with balance > 0 but not deployed.

ClementWalter: Also agree on the need for a logger

crates/core/tests/client.rs
ClementWalter: it's not overkill!

ClementWalter (dismissed):

scripts/PlainOpcodes.s.sol
```diff
@@ -5,7 +5,7 @@ import "forge-std/Script.sol";
 import "kakarot/PlainOpcodes/Counter.sol";
 import "kakarot/PlainOpcodes/PlainOpcodes.sol";
 
-contract EvmsheetScript is Script {
+contract PlainOpcodesScript is Script {
```
ClementWalter, line 8: 🤯

ClementWalter (approved):

Issue #253, opened 2023-07-05
feat: create integration tests on eth-rpc crate that use either Katana, Madara or TestSequencer to make soundness tests automatic


Eikix:
# Feature Request

**Describe the Feature Request**

- create integration tests on eth-rpc crate that use either Katana, Madara or TestSequencer to make soundness tests automatic

Currently, checking that we're moving towards our common goal -> working rpc for ethcc requires lots of [manipulations](https://www.notion.so/kakarotlabs/Deploy-Kaka-swap-on-Madara-1d915119a66843329921bc381c913781). 

It'd be easier to use a eth sdk plugged onto eth-rpc 

**Describe Preferred Solution**

Investigate during one day a clean architecture to create integration tests for eth-rpc that can:
- point to a real starknet RPC where kakarot is deployed (madara or katana, or test sequencer)
- deploy solidity smart contracts
- interact with them using eth sdks (ether-rs, etherjs, etc.)
- assert on calls/ invokes

Eikix:
After a day of investigation, comment down here your findings and propose an architecture for integration tests.

jobez:
tried deploying kakaswap on katana manually and am running into the steps issue:

```
[...]

PR #274, opened 2023-07-10
`run_kakarot_integration` takes a client callback for integration tests

jobez:
Time spent on this PR:

Resolves: #253 

# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

`run_kakarot_integration` tales a callback that has the rpc port as argument

- [ ] Yes
- [X] No

# Additional info

I hit a blocker using forge, will revisit when the types are lined up. I recreated the bug with the ethers client as an initial first test, will iterate a better one

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ readme = "./README.md"
 license = "MIT"
 
 [workspace.dependencies]
+ethers = "2.0"  
 starknet = "0.4.0"
 starknet-crypto = "0.5.1"
 reth-rpc-api = { git = "https://github.com/paradigmxyz/reth.git", rev = "fb710e5" }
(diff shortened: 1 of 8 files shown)


crates/eth-rpc/tests/rpc.rs
```diff
@@ -1013,4 +1015,19 @@ mod tests {
         );
         assert_eq!(U256::from(transaction.block_number.unwrap()), U256::from(13));
     }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 3)]
```
jobez, line 1019: i recreated the flow that was failing when we try to use forge for deploy, here.

so it is meant to fail, but i was able to reproduce a smaller minimal case that i created an issue for, so i can iterate and do a different test

crates/core/src/test_utils/deploy_helpers.rs
```diff
@@ -536,6 +536,45 @@ impl DeployedKakarot {
...
+/// This function generates a `StarknetConfig` instance with `allow_zero_max_fee` set to `true`
+/// and a custom `Environment` object with its `chain_id` set to "SN_GOERLI".
+/// It also sets `invoke_max_steps` and `validate_max_steps` to `2**24` to facilitate
+/// computations associated with the "Kakarot" environment.
+/// Any other configuration fields in `StarknetConfig` and `Environment` are filled with default
+/// values.
+///
+/// This function is intended to provide a convenient way to obtain a Starknet configuration
+/// specifically tailored for testing or running "Kakarot"-centric applications.
+pub fn get_kakarot_starknet_config() -> StarknetConfig {
```
Eikix, line 550: nice!

crates/core/src/test_utils/deploy_helpers.rs
```diff
@@ -536,6 +536,45 @@ impl DeployedKakarot {
     }
 }
 
+/// Returns a `StarknetConfig` instance customized for "Kakarot".
```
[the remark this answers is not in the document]
jobez: will fix, t'was a gpt generated doc weirdness

Cargo.toml
```diff
@@ -40,6 +40,8 @@ forge = { git = "https://github.com/foundry-rs/foundry", branch = "master" }
 # see https://github.com/dojoengine/dojo/issues/563
 dojo-test-utils = { git = "https://github.com/jobez/dojo", branch = "custom-steps-and-get-block-potential-fix"  }
 [patch.crates-io]
+# patched for quantity U256 responses <https://github.com/recmo/uint/issues/224>
+ruint = { git = "https://github.com/paradigmxyz/uint" }
```
jobez, line 44: When we include any rust git dependency, we have to make sure we observe the project's patches, as it can create subtle yet drastic differences in the behavior in the libraries usage!

https://github.com/paradigmxyz/reth/blob/main/Cargo.toml#L74

see:
https://github.com/foundry-rs/foundry/issues/5303#issuecomment-1628822918

Eikix: nice catch, v cursed, i dont understand why projects dont inherit patches?

Eikix: would be cool to find out as a team, philosophically

Issue #277, opened 2023-07-10
bug: eth_getCode is raising when there is no corresponding starknet contract


ClementWalter:
# Bug Report

**Current behavior:**

When querying `eth_getCode`, the RPc would raise with

```
method_call{method="eth_getCode"}: jsonrpsee_core::tracing: send="{\"jsonrpc\":\"2.0\",\"error\":{\"code\":3,\"message\":\"Contract error\"},\"id\":3}"
```

See

https://github.com/kkrt-labs/kakarot-rpc/blob/764c47a703e92cec5faa76e4ead4566be70da559/crates/core/src/client/mod.rs#L116

**Expected behavior:**

The RPC should return `0x` (empty bytes) if there is no code at this address

**Steps to reproduce:**

I used the forge script

```
forge script scripts/PlainOpcodes.s.sol --rpc-url http://127.0.0.1:3030 --broadcast
```

**Related code:**

https://github.com/kkrt-labs/kakarot-rpc/blob/764c47a703e92cec5faa76e4ead4566be70da559/crates/core/src/client/mod.rs#L116

- instead of using `?`, which should handle the error case and return `0x` instead.

PR #280, opened 2023-07-10
Fix getCode should not raise when contract doesn't exist

ClementWalter:
Time spent on this PR: 0.3

Resolves: #277

# Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

When using the RPC with madara, eth_getCode doesn't raise when the queried contract doesn't exist.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -112,8 +112,21 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
             entry_point_selector: BYTECODE,
             calldata: vec![],
         };
+
+        dbg!(&request);
+
         // Make the function call to get the contract bytecode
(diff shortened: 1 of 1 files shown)


crates/core/src/client/mod.rs
```diff
@@ -112,8 +112,21 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
             entry_point_selector: BYTECODE,
             calldata: vec![],
         };
+
+        dbg!(&request);
```
jobez, line 116: secret sneaky dbg!

ClementWalter: you showed me the way

jobez (approved):
lgtm

Issue #273, opened 2023-07-10
feat: refactor starknet provider to sequencergateway provider


Eikix:
# Feature Request

**Describe the Feature Request**

**Describe Preferred Solution**

In eth-rpc initialization, make RPC vs. Sequencer gateway choice a CLI arg or .env arg

PR #283, opened 2023-07-10
feat: add possibility to have sequencer provider or jsonrpc provider

Eikix:
Time spent on this PR: 1day

Resolves: #273 

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

- [ ] Yes
- [x] No

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,16 +5,18 @@ RUST_LOG=debug
 
 # Starknet Environment
 STARKNET_NETWORK=testnet
+## Katana specific configurations
+KATANA_ACCOUNT_ADDRESS=0x03ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0
+KATANA_PRIVATE_KEY=0x0300001800000000300000180000000000030000000000003006001800006600
+## Madara account address and private key
+MADARA_ACCOUNT_ADDRESS=0x3
+MADARA_PRIVATE_KEY=0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7bcd1d
 
 # Kakarot Environment
 KAKAROT_HTTP_RPC_ADDRESS=0.0.0.0:3030
 ## check `./deployments/katana/depolyments.json` after running `make devnet`
 KAKAROT_ADDRESS=
 PROXY_ACCOUNT_CLASS_HASH=0x3010a53967fa04842bcbcb6de8817101f047ef0d074b3eacbe714a3fc42a2eb
-
-# Madara account address and private key
-MADARA_ACCOUNT_ADDRESS=0x3
-MADARA_PRIVATE_KEY=0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7bcd1d
 
 ## configurations for testing
 COMPILED_KAKAROT_PATH=lib/kakarot/build
--- a/.env.example
+++ b/.env.example
@@ -1,8 +1,14 @@
 # Copy this file to .env and fill in the values
-STARKNET_RPC_URL=http://0.0.0.0:5050/rpc
+
+# Rust Environment
 RUST_LOG=debug
+
+# Starknet Environment
+STARKNET_NETWORK=testnet
+
+# Kakarot Environment
 KAKAROT_HTTP_RPC_ADDRESS=0.0.0.0:3030
-# check `./deployments/katana/depolyments.json` after running `make devnet`
+## check `./deployments/katana/depolyments.json` after running `make devnet`
 KAKAROT_ADDRESS=
 PROXY_ACCOUNT_CLASS_HASH=0x3010a53967fa04842bcbcb6de8817101f047ef0d074b3eacbe714a3fc42a2eb
 
@@ -10,9 +16,9 @@ PROXY_ACCOUNT_CLASS_HASH=0x3010a53967fa04842bcbcb6de8817101f047ef0d074b3eacbe714
 MADARA_ACCOUNT_ADDRESS=0x3
 MADARA_PRIVATE_KEY=0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7bcd1d
 
-# configurations for testing
+## configurations for testing
 COMPILED_KAKAROT_PATH=lib/kakarot/build
 
-# An EVM private to define a default EOA for EVM related scripts
-# This default value is Anvil first account private key
+## An EVM private to define a default EOA for EVM related scripts
+## This default value is Anvil first account private key
 EVM_PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
--- a/.env.example
+++ b/.env.example
@@ -14,7 +14,7 @@ MADARA_PRIVATE_KEY=0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7
 
 # Kakarot Environment
 KAKAROT_HTTP_RPC_ADDRESS=0.0.0.0:3030
-## check `./deployments/katana/depolyments.json` after running `make devnet`
+## check `./deployments/katana/deployments.json` after running `make devnet`
 KAKAROT_ADDRESS=
 PROXY_ACCOUNT_CLASS_HASH=0x3010a53967fa04842bcbcb6de8817101f047ef0d074b3eacbe714a3fc42a2eb
 
--- a/crates/core/src/client/config.rs
+++ b/crates/core/src/client/config.rs
@@ -4,31 +4,71 @@ use starknet::providers::jsonrpc::{HttpTransport, JsonRpcTransport};
 use starknet::providers::JsonRpcClient;
 use url::Url;
 
+use super::constants::{KATANA_RPC_URL, MADARA_RPC_URL};
 use super::errors::ConfigError;
 
 fn get_env_var(name: &str) -> Result<String, ConfigError> {
     std::env::var(name).map_err(|_| ConfigError::EnvironmentVariableMissing(name.into()))
 }
 
-#[derive(Default)]
+#[derive(Default, Clone)]
+pub enum Network {
+    #[default]
+    Katana,
+    Madara,
+    Mainnet,
+    Goerli1,
+    Goerli2,
+    Mock,
+}
+
+#[derive(Default, Clone)]
 /// Configuration for the Starknet RPC client.
 pub struct StarknetConfig {
-    /// Rpc url.
-    pub url: String,
+    /// Rpc url. Optional as it can be provided by the environment variable `STARKNET_NETWORK`.
+    pub url: Option<Url>,
     /// Kakarot contract address.
     pub kakarot_address: FieldElement,
     /// Proxy account class hash.
     pub proxy_account_class_hash: FieldElement,
+    /// Additional configuration if the underlying provider is a Sequencer provider.
+    pub network: Network,
 }
 
 impl StarknetConfig {
-    pub fn new(url: String, kakarot_address: FieldElement, proxy_account_class_hash: FieldElement) -> Self {
-        StarknetConfig { url, kakarot_address, proxy_account_class_hash }
+    pub fn new(
+        url: Option<Url>,
+        kakarot_address: FieldElement,
+        proxy_account_class_hash: FieldElement,
+        network: Network,
+    ) -> Self {
+        StarknetConfig { url, kakarot_address, proxy_account_class_hash, network }
     }
 
     /// Create a new `StarknetConfig` from environment variables.
     pub fn from_env() -> Result<Self, ConfigError> {
-        let starknet_rpc = get_env_var("STARKNET_RPC_URL")?;
+        let mut provider_url = None;
+        let starknet_rpc = get_env_var("STARKNET_NETWORK")?;
+        let network = match starknet_rpc.to_lowercase().as_str() {
+            // TODO: Add possibility to set url for katana and madara in env rather than constants.
+            "katana" => {
+                provider_url = Some(Url::parse(KATANA_RPC_URL)?);
+                Network::Katana
+            }
+            "madara" => {
+                provider_url = Some(Url::parse(MADARA_RPC_URL)?);
+                Network::Madara
+            }
+            // TODO: Add possibility to override gateway url for mainnet and testnet.
+            "mainnet" => Network::Mainnet,
+            "goerli1" => Network::Goerli1,
+            "goerli2" => Network::Goerli2,
+            "testnet" => Network::Goerli1,
+            _ => Err(ConfigError::EnvironmentVariableSetWrong(format!(
+                "STARKNET_NETWORK should be either katana, madara, goerli1, goerli2, testnet or mainnet got \
+                 {starknet_rpc}"
+            )))?,
+        };
 
         let kakarot_address = get_env_var("KAKAROT_ADDRESS")?;
         let kakarot_address = FieldElement::from_hex_be(&kakarot_address).map_err(|_| {
@@ -44,7 +84,7 @@ impl StarknetConfig {
             ))
         })?;
 
-        Ok(StarknetConfig::new(starknet_rpc, kakarot_address, proxy_account_class_hash))
+        Ok(StarknetConfig::new(provider_url, kakarot_address, proxy_account_class_hash, network))
     }
 }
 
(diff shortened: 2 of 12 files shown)


crates/eth-rpc/src/main.rs
```diff
@@ -16,17 +17,61 @@ async fn main() -> Result<()> {
...
-    println!("RPC Server running on {url}...");
+            println!("RPC Server running on {url}...");
 
-    server_handle.stopped().await;
+            server_handle.stopped().await;
+        }
+        Network::Mock => {
+            return Err(eyre::eyre!("Cannot run RPC server with mock network"));
+        }
+    }
```
danilowhk, line 74: There seems to be some redundency, maybe we could only initialize the provider within the match and kakarot_client, server_addr, server_handle and url after the match

Eikix: I tried but didn't manage to.
KakarotClient<P> complains when the P is not known at compile time. In the match arms, there could be a JsonRpcClient or a SequencerGatewayProvider.

crates/core/src/client/config.rs
```diff
@@ -4,31 +4,71 @@ use starknet::providers::jsonrpc::{HttpTransport, JsonRpcTransport};
...
+        proxy_account_class_hash: FieldElement,
+        network: Network,
+    ) -> Self {
+        StarknetConfig { url, kakarot_address, proxy_account_class_hash, network }
     }
 
     /// Create a new `StarknetConfig` from environment variables.
     pub fn from_env() -> Result<Self, ConfigError> {
-        let starknet_rpc = get_env_var("STARKNET_RPC_URL")?;
+        let mut provider_url = None;
```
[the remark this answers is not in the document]
Eikix: ok done!

.env.example
```diff
@@ -1,14 +1,30 @@
 # Copy this file to .env and fill in the values
-STARKNET_RPC_URL=http://0.0.0.0:5050/rpc
+
+# Rust Environment
 RUST_LOG=debug
+
+# Starknet Environment
+STARKNET_NETWORK=testnet
+### Starknet provider urls:
+# {
```
ClementWalter, line 9: This is already hardcoded in starknet rs, see https://github.com/xJonathanLEI/starknet-rs/blob/35c287e1a06e6ab68447f5f0b9df53f910960f57/starknet-providers/src/sequencer/mod.rs#L114-L136

crates/eth-rpc/src/main.rs
```diff
@@ -15,17 +16,61 @@ async fn main() -> Result<()> {
     tracing_subscriber::FmtSubscriber::builder().with_env_filter(filter).finish().try_init()?;
 
     let starknet_config = StarknetConfig::from_env()?;
+
     let rpc_config = RPCConfig::from_env()?;
 
-    let provider = JsonRpcClientBuilder::with_http(&starknet_config).unwrap().build();
-    let kakarot_client = KakarotClient::new(starknet_config, provider);
+    // This match creates redundancy but it appears KakarotClient<P> complains about the
```
ClementWalter, line 22: maybe use a `Box<dyn Provider>` somewhere?

crates/core/src/client/config.rs
```diff
@@ -4,31 +4,64 @@ use starknet::providers::jsonrpc::{HttpTransport, JsonRpcTransport};
...
+        proxy_account_class_hash: FieldElement,
+        network: Network,
+    ) -> Self {
+        StarknetConfig { url, kakarot_address, proxy_account_class_hash, network }
     }
 
     /// Create a new `StarknetConfig` from environment variables.
     pub fn from_env() -> Result<Self, ConfigError> {
-        let starknet_rpc = get_env_var("STARKNET_RPC_URL")?;
+        let starknet_rpc = get_env_var("STARKNET_NETWORK")?;
```
ClementWalter, line 50: I don't understand this restriction to a list of target. I think that the config should be instead

```
{
    pub provider_url: "mainnet" | "testnet" | "testnet2" | Url,
}
```

In the `new` we can match against the provider_url and instantiate a sequencer or rpc provider accordingly

In the `from_env` we just take this rpc from end and call `new`

crates/core/src/client/mod.rs
```diff
@@ -315,7 +315,7 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
 
                     TransactionReceipt {
                         transaction_hash,
-                        transaction_index: None,
+                        transaction_index: Some(U256::ZERO),
```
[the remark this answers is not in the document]
jobez: yep*. not sure if its a good move to merge these changes until we are out of an explorational phase. could be overfitting to just forge.

<details>
*it will be on both madara and katana, as it relates to how the ethers client parses our response. 
</details>

.env.example
```diff
@@ -1,14 +1,30 @@
...
+# }
+###
+# Katana specific configurations
+KATANA_ACCOUNT_ADDRESS=0x03ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0
+KATANA_PRIVATE_KEY=0x0300001800000000300000180000000000030000000000003006001800006600
+
+# Kakarot Environment
 KAKAROT_HTTP_RPC_ADDRESS=0.0.0.0:3030
-# check `./deployments/katana/depolyments.json` after running `make devnet`
+## check `./deployments/katana/depolyments.json` after running `make devnet`
```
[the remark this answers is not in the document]
Eikix: just hierarchy in paragraphs

crates/eth-rpc/src/main.rs
Eikix: I'll try

crates/core/src/client/config.rs
Eikix: Alright, I'll try like this, 
The way to do this (union types) is through an enum! It's actually a great idea, thanks for the tip

crates/eth-rpc/src/main.rs
Eikix: ```the trait `starknet::providers::Provider` cannot be made into an object
`starknet::providers::Provider` cannot be made into an objectrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B19%5D?19#file%3A%2F%2F%2FUsers%2Feliastazartes%2Fcode%2Ftheodo%2Fkakarot-rpc%2Fcrates%2Feth-rpc%2Fsrc%2Fmain.rs)
provider.rs(20, 14): for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
```

Comment (Eikix) [reply crates/eth-rpc/src/main.rs]: proposed a fix

File: crates/core/src/client/config.rs (hunk 5 above)
Comment (Eikix) [reply crates/core/src/client/config.rs]: proposed a fix

**Review by jobez — COMMENTED** [review]
a couple wish list dev-ux things re: providers that may be outside the scope of this PR, but can build upon it 🌈 in the future 🌟 

- logging for which provider you are using
- an optional dev mode that will check if your KAKAROT_ADDRESS is deployed yet, warn/panic if not

otherwise, good work!

**Comment by Eikix** [comment]
> a couple wish list dev-ux things re: providers that may be outside the scope of this PR, but can build upon it 🌈 in the future 🌟
> 
> * logging for which provider you are using
> * an optional dev mode that will check if your KAKAROT_ADDRESS is deployed yet, warn/panic if not
> 
> otherwise, good work!

Interesting devX life improvements! I suggest we open an issue and add onto it

File: crates/core/src/client/config.rs [hunk 8]
```diff
@@ -4,31 +4,56 @@ use starknet::providers::jsonrpc::{HttpTransport, JsonRpcTransport};
...
+    Goerli2,
+    ProviderUrl(Url),
+}
+
+#[derive(Default, Clone)]
 /// Configuration for the Starknet RPC client.
 pub struct StarknetConfig {
-    /// Rpc url.
-    pub url: String,
+    /// Additional configuration if the underlying provider is a Sequencer provider.
```
ClementWalter, line 28: is this docstring still accurate?

crates/core/src/client/config.rs
```diff
@@ -4,31 +4,56 @@ use starknet::providers::jsonrpc::{HttpTransport, JsonRpcTransport};
...
     /// Kakarot contract address.
     pub kakarot_address: FieldElement,
     /// Proxy account class hash.
     pub proxy_account_class_hash: FieldElement,
 }
 
 impl StarknetConfig {
-    pub fn new(url: String, kakarot_address: FieldElement, proxy_account_class_hash: FieldElement) -> Self {
-        StarknetConfig { url, kakarot_address, proxy_account_class_hash }
+    pub fn new(network: Network, kakarot_address: FieldElement, proxy_account_class_hash: FieldElement) -> Self {
```
ClementWalter, line 37: do we need to take ownership of the args? (maybe more general quesiton, but I don't see if we have any general guideline regarding this)

crates/core/src/client/config.rs
```diff
@@ -4,31 +4,56 @@ use starknet::providers::jsonrpc::{HttpTransport, JsonRpcTransport};
...
 
     /// Create a new `StarknetConfig` from environment variables.
     pub fn from_env() -> Result<Self, ConfigError> {
-        let starknet_rpc = get_env_var("STARKNET_RPC_URL")?;
+        let network = get_env_var("STARKNET_NETWORK")?;
+        let network = match network.to_lowercase().as_str() {
+            // TODO: Add possibility to set url for katana and madara in env rather than constants.
+            "katana" => Network::Katana,
+            "madara" => Network::Madara,
+            // TODO: Add possibility to override gateway url for mainnet and testnet.
```
ClementWalter, line 48: I think that it's not required, and I don't see any situation where it could be used. There is only one gateway per network

crates/core/src/client/config.rs
```diff
@@ -4,31 +4,56 @@ use starknet::providers::jsonrpc::{HttpTransport, JsonRpcTransport};
...
+            "mainnet" => Network::Mainnet,
+            "goerli1" => Network::Goerli1,
+            "goerli2" => Network::Goerli2,
+            "testnet" => Network::Goerli1,
+            _ => Err(ConfigError::EnvironmentVariableSetWrong(format!(
+                "STARKNET_NETWORK should be either katana, madara, goerli1, goerli2, testnet or mainnet got {network}"
```
ClementWalter, line 54: any RPC url no?

crates/core/src/client/config.rs
```diff
@@ -4,31 +4,56 @@ use starknet::providers::jsonrpc::{HttpTransport, JsonRpcTransport};
...
+        StarknetConfig { network, kakarot_address, proxy_account_class_hash }
     }
 
     /// Create a new `StarknetConfig` from environment variables.
     pub fn from_env() -> Result<Self, ConfigError> {
-        let starknet_rpc = get_env_var("STARKNET_RPC_URL")?;
+        let network = get_env_var("STARKNET_NETWORK")?;
+        let network = match network.to_lowercase().as_str() {
+            // TODO: Add possibility to set url for katana and madara in env rather than constants.
+            "katana" => Network::Katana,
```
ClementWalter, line 46: ~what is the need for having katana and madara? what does this info bring compared to a plain rpc url?~ ok just swa below, to avoid to type the default url. From our xp in kakarot, if we want to store some config per network, we need to have a object per network, and not for each param (here below the url), one entry per network

crates/eth-rpc/src/main.rs
```diff
@@ -2,10 +2,22 @@ use dotenv::dotenv;
...
 use kakarot_rpc::config::RPCConfig;
 use kakarot_rpc::run_server;
-use kakarot_rpc_core::client::config::{JsonRpcClientBuilder, StarknetConfig};
+use kakarot_rpc_core::client::config::{JsonRpcClientBuilder, Network, StarknetConfig};
 use kakarot_rpc_core::client::KakarotClient;
+use starknet::providers::jsonrpc::HttpTransport;
+use starknet::providers::{JsonRpcClient, SequencerGatewayProvider};
 use tracing_subscriber::util::SubscriberInitExt;
 
+enum KakarotProvider {
```
ClementWalter, line 11: `StarknetProvider` maybe?

crates/eth-rpc/src/main.rs
```diff
@@ -15,12 +27,45 @@ async fn main() -> Result<()> {
...
+        KakarotProvider::JsonRpcClient(provider) => {
+            KakarotClientType::RpcClient(KakarotClient::new(starknet_config, provider))
+        }
+        KakarotProvider::SequencerGatewayProvider(provider) => {
+            KakarotClientType::GatewayClient(KakarotClient::new(starknet_config, provider))
+        }
+    };
+
+    let (server_addr, server_handle) = match kakarot_client {
+        KakarotClientType::GatewayClient(kakarot_client) => run_server(Box::new(kakarot_client), rpc_config).await?,
```
ClementWalter, line 65: why is the `match` required? can't we use directly `run_server(Box::new(kakarot_client), rpc_config).await?`?

crates/core/src/client/config.rs
Eikix: good question, I'm not sure, maybe we could investigate this further in a separate issue, as it has long reaching ramifications in the code base

crates/core/src/client/config.rs
Eikix: good catch

crates/core/src/client/config.rs
Eikix: but it shouldn't be here then, mb!

crates/core/src/client/config.rs
Eikix: what do you mean?

crates/eth-rpc/src/main.rs
Eikix: because run_server takes in a kakarot_client type, not an enum wrapping this type

Eikix: we could implement on the enum an .into_inner() function that returns the inner type, using a match

crates/eth-rpc/src/main.rs
Eikix: ok

crates/core/src/client/config.rs
Eikix: fixed

crates/core/src/client/config.rs
ClementWalter: like in kakarot we used to have

```
RPC_URLS={
    "testnet": "https//",
    "devnet": "https://",
}

EXPLORER_URL = {
    "testnet": "testnet.starkscan"
    "devnet": "devnet.starkscan"
}

# ...
```

and we changed to have

```
NETWORKS = {
    testnet: {
        rpc_url?:
        gateway?:
        explorer:
    },
    devnet: {
# ..
}
```

ie a list of networks and not a list par parameter

Eikix: Why do we need this here though?

crates/eth-rpc/src/main.rs
```diff
@@ -15,12 +27,45 @@ async fn main() -> Result<()> {
     tracing_subscriber::FmtSubscriber::builder().with_env_filter(filter).finish().try_init()?;
 
     let starknet_config = StarknetConfig::from_env()?;
+
     let rpc_config = RPCConfig::from_env()?;
 
-    let provider = JsonRpcClientBuilder::with_http(&starknet_config).unwrap().build();
-    let kakarot_client = KakarotClient::new(starknet_config, provider);
+    let provider: StarknetProvider = match &starknet_config.network {
```
ClementWalter, line 33: `let starknet_provider: StarknetProvider = `

i find it difficult to follow what is what generally speaking, and that prepending with Eth / Starknet helps

Eikix: fixed

crates/eth-rpc/src/main.rs
Eikix: I agree! I just don't know yet how to do it

Eikix: Let's open an issue!

crates/core/src/client/config.rs
ClementWalter: we have here a match, then we load KATANA_URL and MADARA_URL at some other places. Maybe we want to `serde` a json config file. not required for this PR though

Eikix: I see!

Issue #291, opened 2023-07-11
feat: implement `KakarotRPC` struct, to make adding more RPC methods with various concerns easy.


bajpai244:
# Feature Request

We should implement a new struct `KakarotRpc`, that should allow merging methods from various traits easily, and should allow us to expose them via a single server and port.

## Why do we need this?

When looking into the [following](https://github.com/kkrt-labs/kakarot-rpc/blob/b6fa860d73bd85ef69b690daa3eadf1ee76b8961/crates/eth-rpc/src/lib.rs#L38) line of `run_server` function, we use an `into_rpc` method, this is method is created by a macro expansion, and this is why, we can only have one trait with `[#rpc]` attribute for our `KakarotEthRpc` struct.

As we move ahead in the RPC, we will be adding more and more RPC methods that might not be part of the official spec but would be implemented by various other execution clients, which will require us to implement them to support tooling. We would want there to be a separation of concerns between these Apis and hence would need different traits defining them { for example we need one new trait for #248 issue }.

This refactor will make adding new traits with RPC handlers easier. We can create a new trait { like `Web3Api`,  `Tracing` }, and then just update `KakarotRpc` struct to have them as added as well.

bajpai244:
If everyone agrees on this, then I can take this, and my pr for #248 is also dependent on this.

PR #293, opened 2023-07-11
feat: add kakarotRpc struct to better manager our RPC implementation

bajpai244:
In this proposed solution, every time we want to add an API, we put it in the API directory, then we implement a server for the API, and then we can update our KakarotRPC struct.

KakarotRPC struct can then expose a single `RpcModule`, which can run all these methods from these various traits on the same server and port, we are able to separate concerns, at the same time during runtime able to run them all under a single server.

solves #291


crates/core/src/client/config.rs
```diff
@@ -48,6 +48,16 @@ impl StarknetConfig {
     }
 }
 
+impl Clone for StarknetConfig {
```
Eikix, line 51: What is the purpose of this manual implementation of clone?

- I suspect you did this because you can't just derive Clone for StarknetConfig?
- Is there a workaround for this? Namely, use references when needed rather than clones or functions that consume the ownership?

crates/eth-rpc/src/rpc.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use crate::api::eth_api::EthApiServer;
+use crate::api::kakarot_custom_api::KakarotCustomApiServer;
+use crate::servers::eth_rpc;
+use crate::servers::eth_rpc::KakarotEthRpc;
+use crate::servers::kakarot_custom_rpc::{self, KakarotCustomRpc};
+
+pub struct KakarotRpc {
+    eth_rpc_module: RpcModule<KakarotEthRpc<JsonRpcClient<HttpTransport>>>,
+    kakarot_custom_rpc_module: RpcModule<KakarotCustomRpc<JsonRpcClient<HttpTransport>>>,
```
Eikix, line 15: Same here, especially if they use the same type (`JsonRpcClient<HttpTransport>`), it could be passed as a generic to `KakarotRpc`

crates/eth-rpc/src/rpc.rs
```diff
@@ -0,0 +1,42 @@
...
+            eth_rpc::KakarotEthRpc::new(Box::new(KakarotRpc::get_kakarot_client(starknet_config.clone()))).into_rpc();
+        let kakarot_custom_rpc_module =
+            kakarot_custom_rpc::KakarotCustomRpc::new(Box::new(KakarotRpc::get_kakarot_client(starknet_config)))
+                .into_rpc();
+
+        Self { eth_rpc_module, kakarot_custom_rpc_module }
+    }
+
+    pub fn get_rpc_module(&self) -> RpcModule<()> {
+        let mut rpc_module = RpcModule::new(());
```
Eikix, line 30: is there a slightly cleaner way to do this merging?

crates/eth-rpc/src/rpc.rs
```diff
@@ -0,0 +1,42 @@
...
+
+        rpc_module.merge(self.eth_rpc_module.clone()).unwrap();
```
Eikix, line 32: why do we unwrap here? is there a way around it?

crates/eth-rpc/src/rpc.rs
```diff
@@ -0,0 +1,42 @@
...
+        let eth_rpc_module =
+            eth_rpc::KakarotEthRpc::new(Box::new(KakarotRpc::get_kakarot_client(starknet_config.clone()))).into_rpc();
+        let kakarot_custom_rpc_module =
+            kakarot_custom_rpc::KakarotCustomRpc::new(Box::new(KakarotRpc::get_kakarot_client(starknet_config)))
+                .into_rpc();
+
+        Self { eth_rpc_module, kakarot_custom_rpc_module }
+    }
+
+    pub fn get_rpc_module(&self) -> RpcModule<()> {
```
Eikix, line 29: can we add rustdoc comments
akin to what @jobez did in his recent PRs:)?

just explain the general flow of things

crates/eth-rpc/src/rpc.rs
```diff
@@ -0,0 +1,42 @@
...
+        let mut rpc_module = RpcModule::new(());
+
+        rpc_module.merge(self.eth_rpc_module.clone()).unwrap();
+        rpc_module.merge(self.kakarot_custom_rpc_module.clone()).unwrap();
+
+        rpc_module
+    }
+
+    fn get_kakarot_client(starknet_config: StarknetConfig) -> KakarotClient<JsonRpcClient<HttpTransport>> {
+        let provider = JsonRpcClientBuilder::with_http(&starknet_config).unwrap().build();
```
Eikix, line 39: why unwrap? was this already there?

crates/eth-rpc/src/servers/kakarot_custom_rpc.rs
```diff
@@ -0,0 +1,27 @@
+use jsonrpsee::core::{async_trait, RpcResult as Result};
+use kakarot_rpc_core::client::api::KakarotEthApi;
+use kakarot_rpc_core::models::balance::TokenBalances;
+use reth_primitives::Address;
+use starknet::providers::Provider;
+
+use crate::api::kakarot_custom_api::KakarotCustomApiServer;
+
+/// The RPC module for the Ethereum protocol required by Kakarot.
+pub struct KakarotCustomRpc<P: Provider + Send + Sync> {
```
Eikix, line 10: I think I got this API from alchemy
https://docs.alchemy.com/reference/ethereum-api-endpoints

let's rename it to give them many thanks

crates/eth-rpc/src/rpc.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::servers::eth_rpc::KakarotEthRpc;
+use crate::servers::kakarot_custom_rpc::{self, KakarotCustomRpc};
+
+pub struct KakarotRpc {
+    eth_rpc_module: RpcModule<KakarotEthRpc<JsonRpcClient<HttpTransport>>>,
+    kakarot_custom_rpc_module: RpcModule<KakarotCustomRpc<JsonRpcClient<HttpTransport>>>,
+}
+
+impl KakarotRpc {
+    pub fn new(starknet_config: StarknetConfig) -> Self {
```
Eikix, line 19: a problem I see here is that we're using 2 different `KakarotClient` instances, one for each API, but they're the same client! 
It'll end up being quite inefficient

Possible workarounds:
Having Arc clones of KakarotClient passed around?

You'll have to investigate if that doesn't work out -> our main philosophical challenge here is: each api will probably point to core crate, thus kakarot client, how can we not duplicate kakarot client instances for each separate api

Eikix (changes requested):
great PR! going to give us lots of freedom to add new APIs to our json rpc server

let me know if comments are clear

bajpai244:
> great PR! going to give us lots of freedom to add new APIs to our json rpc server
> 
> let me know if comments are clear

@Eikix thanks for the review, working on these comments by resolving one at a time.

I would also consider to watch for PRs we are going to merge before this, to see if there can be big rebase issues.

On it ⚡

crates/eth-rpc/src/rpc.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub fn get_rpc_module(&self) -> RpcModule<()> {
+        let mut rpc_module = RpcModule::new(());
+
+        rpc_module.merge(self.eth_rpc_module.clone()).unwrap();
+        rpc_module.merge(self.kakarot_custom_rpc_module.clone()).unwrap();
+
+        rpc_module
+    }
+
+    fn get_kakarot_client(starknet_config: StarknetConfig) -> KakarotClient<JsonRpcClient<HttpTransport>> {
```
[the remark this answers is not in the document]
bajpai244: Done.

crates/eth-rpc/src/rpc.rs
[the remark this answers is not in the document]
bajpai244: Done.

crates/eth-rpc/src/servers/kakarot_custom_rpc.rs
bajpai244: Makes sense. Resolved.

Also, then it means we are not completely implementing the Token Api, we just have one method, should I make an issue for this?

image

crates/core/src/client/config.rs
ClementWalter: Actually I was able to compile

```
#[derive(Default, Clone)]
/// Configuration for the Starknet RPC client.
pub struct StarknetConfig {
    /// Rpc url.
    pub url: String,
    /// Kakarot contract address.
    /// Proxy account class hash.
}
```

crates/eth-rpc/src/api/kakarot_custom_api.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[rpc(server)]
+#[async_trait]
+pub trait KakarotCustomApi {
+    #[method(name = "kakarot_getTokenBalances")]
+    async fn token_balances(
+        &self,
+        address: Address,
+        contract_addresses: Vec<Address>,
+    ) -> jsonrpsee::core::RpcResult<TokenBalances>;
```
ClementWalter, line 13: can we keep the same semantic as in the other trait, ie import 

```
use jsonrpsee::core::RpcResult as Result;
```

and return

```
Result<TokenBalances>
```

crates/eth-rpc/src/servers/kakarot_custom_rpc.rs
ClementWalter: imho these `alchemy_` endpoints are very low priority for us. It is was good catch to realize that we didn't serve what we thought we were serving, and we definitely could define one trait per prefix even in the "standard" et json rpc spec (`eth_`, `net_`, `web3_`). This `alchemy_` RPC is very low priority for us, so why not an issue, but with "nice to have", and "low priority" tags!

crates/eth-rpc/src/api/kakarot_custom_api.rs
```diff
@@ -0,0 +1,14 @@
+use jsonrpsee::proc_macros::rpc;
+use kakarot_rpc_core::models::balance::TokenBalances;
+use reth_primitives::Address;
+
+#[rpc(server)]
+#[async_trait]
+pub trait KakarotCustomApi {
```
ClementWalter, line 7: so this would be `AlchemyApi`, and file renamed accordingly

crates/eth-rpc/src/api/kakarot_custom_api.rs
```diff
@@ -0,0 +1,14 @@
...
+    #[method(name = "kakarot_getTokenBalances")]
```
ClementWalter, line 8: isn't there a way (I think there is) to automatically prepend the `alchemy_` (here `kakarot_`) to all `#[methods()]` (maybe create an issue and let's update this for the eth api as well)

crates/eth-rpc/src/rpc.rs
```diff
@@ -0,0 +1,42 @@
...
+use starknet::providers::JsonRpcClient;
+
+use crate::api::eth_api::EthApiServer;
+use crate::api::kakarot_custom_api::KakarotCustomApiServer;
+use crate::servers::eth_rpc;
+use crate::servers::eth_rpc::KakarotEthRpc;
+use crate::servers::kakarot_custom_rpc::{self, KakarotCustomRpc};
+
+pub struct KakarotRpc {
+    eth_rpc_module: RpcModule<KakarotEthRpc<JsonRpcClient<HttpTransport>>>,
```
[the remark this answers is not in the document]
ClementWalter: shouldn't it be `Provider` instead?

ClementWalter (commented):
This PR goes to me into the right direction. I'm still confused by our

```
impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P>
```

but this for clarifies already the structure of the RPC

crates/eth-rpc/src/rpc.rs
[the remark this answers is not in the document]
bajpai244: @ClementWalter agreed, @Eikix I have pushed new changes which makes the whole struct generic over the provider.

image

crates/eth-rpc/src/rpc.rs
bajpai244: This is also now generic over `Provider`.

image

crates/eth-rpc/src/rpc.rs
bajpai244: We no more need this function with the new changes I have pushed.

crates/eth-rpc/src/servers/kakarot_custom_rpc.rs
bajpai244: @ClementWalter , noted, whenI make the issue, I will make sure to use these tags.

crates/eth-rpc/src/rpc.rs
bajpai244: Resolved with the new code changes I pushed.

image

crates/core/src/client/config.rs
bajpai244: So, we no more need this, also yeah I could have simply derived, but I guess my Rust noobness made me implement the trait manually! 

TLDR;

We no longer need clone.

crates/eth-rpc/src/rpc.rs
jobez: gpt does a decent job at this, but give yourself time to read through it yourself before you push

crates/eth-rpc/src/servers/kakarot_custom_rpc.rs
bajpai244: It has been renamed accordingly -> 
image

bajpai244:
> > What is the purpose of this manual implementation of clone?
> 
> also wondering when we use borrow and when we take ownership of the args in out functions

We no longer need it.

crates/eth-rpc/src/api/kakarot_custom_api.rs
bajpai244: Done.

crates/eth-rpc/src/api/kakarot_custom_api.rs
bajpai244: Yes, we can create namespaces and do that, I am planning a next PR after it for that,

because:
- We would have to do that in EthApi as well
-  I can see net_version doesn't start with "eth"
-  which means we probably would have to refactor a new trait for it, and I think in that case I would prefer to do it in another PR, lemme know what you think. 
image

crates/eth-rpc/src/api/kakarot_custom_api.rs
bajpai244: Yes, we can create namespaces and do that, I am planning a next PR after it for that,

because:
- We would have to do that in EthApi as well
-  I can see net_version doesn't start with "eth"
-  which means we probably would have to refactor a new trait for it, and I think in that case I would prefer to do it in another PR, lemme know what you think. 
image

crates/eth-rpc/src/rpc.rs
bajpai244: I am returning a Result now { I was hacking fast on the PR, and then pulled it out from draft on daily without remembering about it!  }

image

crates/eth-rpc/src/main.rs
```diff
@@ -17,7 +21,11 @@ async fn main() -> Result<()> {
     let starknet_config = StarknetConfig::from_env()?;
     let rpc_config = RPCConfig::from_env()?;
 
-    let kakarot_rpc = KakarotRpc::new(starknet_config);
+    let provider = JsonRpcClientBuilder::with_http(&starknet_config).unwrap().build();
```
Eikix, line 24: no need to unwrap here I think

crates/eth-rpc/src/main.rs
```diff
@@ -17,7 +21,11 @@ async fn main() -> Result<()> {
...
+    let kakarot_client: Arc<Box<dyn KakarotEthApi<_>>> =
+        Arc::new(Box::new(KakarotClient::new(starknet_config, provider)));
```
Eikix, line 26: quick sanity check: is an Arc of a Box considered "ok" in Rust?

crates/eth-rpc/src/rpc.rs
```diff
@@ -1,41 +1,34 @@
...
+    token_rpc_module: RpcModule<TokenRpc<P>>,
 }
 
-impl KakarotRpc {
-    pub fn new(starknet_config: StarknetConfig) -> Self {
-        let eth_rpc_module =
-            eth_rpc::KakarotEthRpc::new(Box::new(KakarotRpc::kakarot_client(starknet_config.clone()))).into_rpc();
-        let kakarot_custom_rpc_module =
-            kakarot_custom_rpc::KakarotCustomRpc::new(Box::new(KakarotRpc::kakarot_client(starknet_config))).into_rpc();
+impl<P: Provider + Send + Sync + 'static> KakarotRpc<P> {
```
Eikix, line 19: nice!

crates/eth-rpc/src/rpc.rs
Eikix: get_rpc_module -> rpc_module ("get" idiomatic rust)

and let's add rustdoc

crates/eth-rpc/src/rpc.rs
Eikix: what do you mean?

crates/eth-rpc/src/main.rs
Eikix: ah well this was done before^^ interesting. 
i think we should utilize `eyre` to exit gracefully rather than unwrap

Eikix: if not, we can also use "expect"

Eikix: can be done in another issue, let's just create the issue and tackle this in another PR
great job

Eikix (approved):
lgtm! gg @bajpai244

crates/eth-rpc/src/rpc.rs
bajpai244: Did this refactor which is cleaner, lemme know what your thoughts are.

Took some [inspiration](https://github.com/paradigmxyz/reth/blob/dbafe23cce45917a43f6640d71a0216fe3268428/crates/rpc/rpc-builder/src/lib.rs#L632) from reth, tried to keep our implementation as simple as possible.

image

crates/eth-rpc/src/rpc.rs
```diff
@@ -1,34 +1,49 @@
...
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub enum KakarotRpcModule {
+    Eth,
+    Alchemy,
 }
 
-impl<P: Provider + Send + Sync + 'static> KakarotRpc<P> {
+pub struct KakarotRpcModuleBuilder<P: Provider + Send + Sync + 'static> {
+    modules: HashMap<KakarotRpcModule, Methods>,
+    _phantom: PhantomData<P>,
```
Eikix, line 24: why do we need phantom data?

crates/eth-rpc/src/rpc.rs
```diff
@@ -1,34 +1,49 @@
...
+
+        modules.insert(KakarotRpcModule::Eth, eth_rpc_module.into());
+        modules.insert(KakarotRpcModule::Alchemy, alchemy_rpc_module.into());
 
-        Self { eth_rpc_module, token_rpc_module }
+        Self { modules, _phantom: PhantomData }
     }
 
-    pub fn rpc_module(&self) -> RpcModule<()> {
+    pub fn rpc_module(&self) -> Result<RpcModule<()>, Error> {
```
Eikix, line 40: i guess following the builder pattern, this fn should be name `build` or `build_rpc_module`

crates/eth-rpc/src/lib.rs
```diff
@@ -25,7 +25,7 @@ pub enum RpcError {
 ///
 /// Will return `Err` if an error occurs when running the `ServerBuilder` start fails.
 pub async fn run_server<P: Provider + Send + Sync + 'static>(
```
Eikix, line 27: I believe we should keep the same signature for run_server, and trigger the build beforehand *** C.F below comment

crates/eth-rpc/src/main.rs
```diff
@@ -25,7 +25,7 @@ async fn main() -> Result<()> {
     let kakarot_client: Arc<Box<dyn KakarotEthApi<_>>> =
         Arc::new(Box::new(KakarotClient::new(starknet_config, provider)));
 
-    let kakarot_rpc = KakarotRpc::new(kakarot_client);
+    let kakarot_rpc = KakarotRpcModuleBuilder::new(kakarot_client);
```
Eikix, line 28: > I believe we should keep the same signature for run_server, and trigger the build beforehand *** C.F below comment

We should trigger the build right here

crates/eth-rpc/src/lib.rs
bajpai244: Done.

crates/eth-rpc/src/rpc.rs
bajpai244: makes sense, refactored to `build_rpc_module`

crates/eth-rpc/src/rpc.rs
bajpai244: Without it the compiler will throw error for type parameter `P`

image

crates/eth-rpc/src/main.rs
bajpai244: Gotcha! So that makes two issues we have to create:
- Implement all Alchemy endpoints { low priority }
- Refactor to avoid unwrapping here.

After the merge if you want I can create them, or if you want to then that is also fine.

Eikix:
you got conflicts, rip was close to merge first

crates/eth-rpc/src/main.rs
Eikix: Yes! you can create them sounds good

Eikix (approved):
great job!

PR #295, opened 2023-07-12
changes to allow usage `forge create` and `cast`

jobez:
Time spent on this PR:

Resolves: #<Issue number>

# Pull Request type

- [X] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

I previously justified the changes to the `transaction_index` default return value in TransactionReceipt. 

A similar deserializatior error happens when trying to use forge create like so:

`forge create  "Counter" --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80`

You would see something like:

```
Deserialization Error: invalid type: null, expected a sequence at line 1 column 177. Response: {"baseFeePerGas":["0x1","0x1","0x1","0x1","0x1","0x1","0x1","0x1","0x1","0x1","0x1"],"gasUsedRatio":[0.9,0.9,0.9,0.9,0.9,0.9,0.9,0.9,0.9,0.9],"oldestBlock":"0x2b3","reward":null}
```

So we do a similar change to the default value for the 'reward' field. 

I also included a `src` dimension in the `foundry.yml`, which `forge create` needs to locate contracts.

You can see a manual session I did that shows contract deployment/interaction from the command line: https://gist.github.com/jobez/f1c3f655308d080757569f0c10fb4ba3

- [ ] Yes
- [X] No

### DIFF
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -499,7 +499,7 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
 
         // TODO: transition `reward` hardcoded default out of nearing-demo-day hack and seeing how to
         // properly source/translate this value
-        Ok(FeeHistory { base_fee_per_gas, gas_used_ratio, oldest_block, reward: Some(vec![vec![U256::ZERO]]) })
+        Ok(FeeHistory { base_fee_per_gas, gas_used_ratio, oldest_block, reward: Some(vec![vec![]]) })
     }
 
     /// Returns the estimated gas for a transaction
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -315,6 +315,8 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
 
                     TransactionReceipt {
                         transaction_hash,
+                        // TODO: transition this hardcoded default out of nearing-demo-day hack and seeing how to
+                        // properly source/translate this value
                         transaction_index: Some(U256::ZERO),
                         block_hash,
                         block_number,
@@ -495,6 +497,8 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
         let newest_block = U256::from(newest_block);
         let oldest_block: U256 = if newest_block >= block_count { newest_block - block_count } else { U256::from(0) };
 
+        // TODO: transition `reward` hardcoded default out of nearing-demo-day hack and seeing how to
+        // properly source/translate this value
         Ok(FeeHistory { base_fee_per_gas, gas_used_ratio, oldest_block, reward: Some(vec![vec![U256::ZERO]]) })
     }
 
--- a/crates/core/src/client/mod.rs
+++ b/crates/core/src/client/mod.rs
@@ -495,7 +495,7 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
         let newest_block = U256::from(newest_block);
         let oldest_block: U256 = if newest_block >= block_count { newest_block - block_count } else { U256::from(0) };
 
-        Ok(FeeHistory { base_fee_per_gas, gas_used_ratio, oldest_block, reward: None })
+        Ok(FeeHistory { base_fee_per_gas, gas_used_ratio, oldest_block, reward: Some(vec![vec![U256::ZERO]]) })
     }
 
     /// Returns the estimated gas for a transaction
--- a/foundry.toml
+++ b/foundry.toml
@@ -1,6 +1,7 @@
 [profile.default]
 out = 'lib/kakarot/tests/integration/solidity_contracts/build'
 libs = ['lib']
+src = "lib/kakarot/tests/integration/solidity_contracts"
 eth_rpc_url = "localhost"
 legacy = true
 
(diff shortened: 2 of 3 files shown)


crates/core/src/client/mod.rs
```diff
@@ -315,7 +315,7 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
 
                     TransactionReceipt {
                         transaction_hash,
-                        transaction_index: None,
+                        transaction_index: Some(U256::ZERO),
```
jobez, line 318: necessary for `forge script` usage, otherwise deserialization error

crates/core/src/client/mod.rs
```diff
@@ -495,7 +495,7 @@ impl<P: Provider + Send + Sync> KakarotEthApi<P> for KakarotClient<P> {
         let newest_block = U256::from(newest_block);
         let oldest_block: U256 = if newest_block >= block_count { newest_block - block_count } else { U256::from(0) };
 
-        Ok(FeeHistory { base_fee_per_gas, gas_used_ratio, oldest_block, reward: None })
+        Ok(FeeHistory { base_fee_per_gas, gas_used_ratio, oldest_block, reward: Some(vec![vec![U256::ZERO]]) })
```
jobez, line 498: necessary for `forge create` usage, otherwise...

🥁 🥁 🥁 🥁 

you guessed it! deserialization errors

(i wonder if theres a forked library we aren't including that would automatically handle all these weird edge cases for us--there has to be, right?)

foundry.toml
```diff
@@ -1,6 +1,7 @@
 [profile.default]
 out = 'lib/kakarot/tests/integration/solidity_contracts/build'
 libs = ['lib']
+src = "lib/kakarot/tests/integration/solidity_contracts"
```
jobez, line 4: necessary to be able to just have simple contract names for `forge create`

jobez (commented):
minimal changes for using foundry tooling to deploy and interact with deployed contracts

crates/core/src/client/mod.rs
ClementWalter: I think that we need to compute the actual transaction index from the filtered block, ie:
- get starknet block
- filter eth tx
- get index of this tx in the filtered block

crates/core/src/client/mod.rs
ClementWalter: Why do we expect that None is a valid response?

ClementWalter:
I posted decoded error in your gist. I think that this PR deserve a more precise handling ot the index, or a TODO and a linked issue

jobez:
I agree!

crates/core/src/client/mod.rs
Eikix: is the index the index of the tx in the whole block or the index of the tx in the subblock?

crates/core/src/client/mod.rs
Eikix: Because we don't have rewards on Starknet I think

Eikix:
> I posted decoded error in your gist. I think that this PR deserve a more precise handling ot the index, or a TODO and a linked issue

I vote for TODO and linked issue to respect boundaries of scopes!

ClementWalter:
> > I posted decoded error in your gist. I think that this PR deserve a more precise handling ot the index, or a TODO and a linked issue

I vote for TODO and linked issue to respect boundaries of scopes! 

ok for me

crates/core/src/client/mod.rs
ClementWalter: I don't really know when this is used, but I'd say that we filter out the index and return the EVM index because clients may request tx count, then tx by index?

crates/core/src/client/mod.rs
ClementWalter: what about an empty `vec![]` I mean, it's not obvious that `None` could be a possible option

ClementWalter: Is it possible that returning `[[0]]` be interpreted as empty block?

jobez: behold, the power of strong typing across multiple repos:

https://github.com/gakonst/ethers-rs/blob/master/ethers-core/src/types/fee.rs#L18

```
#[derive(Deserialize, Serialize, Debug, Clone)]
#[serde(rename_all = "camelCase")]
pub struct FeeHistory {
    pub base_fee_per_gas: Vec<U256>,
    pub gas_used_ratio: Vec<f64>,
    #[serde(deserialize_with = "deserialize_stringified_numeric")]
    /// oldestBlock is returned as an unsigned integer up to geth v1.10.6. From
    /// geth v1.10.7, this has been updated to return in the hex encoded form.
    /// The custom deserializer allows backward compatibility for those clients
    /// not running v1.10.7 yet.
    pub oldest_block: U256,
    /// An (optional) array of effective priority fee per gas data points from a single block. All
    /// zeroes are returned if the block is empty.
    #[serde(default)]
    pub reward: Vec<Vec<U256>>,
}

```

ClementWalter: lol, writing "optional" in docstring, but non making it `Option` in the code. Can we use an empty `vec<vec<>>` instead?

ClementWalter (changes requested):
it seems like gh bugs and I need to review again and again, so I'm just asking if it's possible to use an empty vec instead of 0 in the rewards

crates/core/src/client/mod.rs
jobez: trying rn, will see if it compiles

jobez:
> it seems like gh bugs and I need to review again and again, so I'm just asking if it's possible to use an empty vec instead of 0 in the rewards

u got it

ClementWalter (approved):

PR #300, opened 2023-07-12
Stop deploying Kakarot in run-dev

ClementWalter:
I am suggesting to stop deploying kakarot when running `make run-dev` because I often keep a madara open, with
once-for-all deployed Kakarot, but I do kill and restart the RPC.

wdyt?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ build:
 run: 
 	RUST_LOG=info cargo run -p kakarot-rpc
 
-run-dev: deploy-kakarot
+run-dev:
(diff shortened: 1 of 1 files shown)


jobez:
yeah locally i have a dev2 that just starts without deploy

ClementWalter:
> yeah locally i have a dev2 that just starts without deploy

so you keep an uncommited file in your workspace? and your workspace is never all clean?

Eikix (approved):
agreed it's v painful for me too!

Issue #248, opened 2023-07-05
feat: Add web3_clientVersion to the RPC


ClementWalter:
# Feature Request

**Describe the Feature Request**

Add https://ethereum.org/en/developers/docs/apis/json-rpc/#web3_clientversion

**Additional Context**

Might not seem so important but some clients like web3 are polling it, see https://github.com/kkrt-labs/kaka-swap/pull/8/files

danilowhk:
I can take this one

jobez:
might be helpful: https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc/src/web3.rs#L29

bajpai244:
I talked with @danilowhk and I am working on it.

PR #308, opened 2023-07-13
feat: add Web3Api

bajpai244:
Add the Web3Api to Kakarot RPC { endpoints that are prefixed "web3_" }

Resolves: #248 

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

Web3Api has been added to Kakarot.

- [] Yes
- [x] No


crates/eth-rpc/src/servers/web3_rpc.rs
```diff
@@ -0,0 +1,27 @@
...
+    }
+}
+
+#[async_trait]
+impl Web3ApiServer for Web3Rpc {
+    fn client_version(&self) -> Result<String> {
+        Ok(String::from(KAKAROT_CLIENT_VERSION))
+    }
+
+    fn sha3(&self, input: Bytes) -> Result<H256> {
```
bajpai244, line 23: For now we just return Keccak256, but we should in future investigate how tools use this endpoint, and according to that what hash should we actually returns { given we use a mix of hashing algorithms afaik }

crates/core/src/client/constants.rs
```diff
@@ -5,6 +5,9 @@ use starknet::core::types::FieldElement;
 /// CHAIN_ID = KKRT (0x4b4b5254) in ASCII
 pub const CHAIN_ID: u64 = 1_263_227_476;
 
+// TODO: Automate changing this version as we do future releases
```
bajpai244, line 8: Added a todo, so that in future we can automate this &str being updates as we bump new versions.

bajpai244: It should probably take the version from `Cargo.toml` of the main RPC workspace.

Cargo.toml
```diff
@@ -3,6 +3,7 @@ members = ["crates/eth-rpc", "crates/core"]
 resolver = "2"
 
 [workspace.package]
+version = "0.1.0"
```
bajpai244, line 6: I think we should manage our version from workspace, and all other packages can share the same version.

Lemme know if you think otherwise.

crates/eth-rpc/src/servers/web3_rpc.rs
```diff
@@ -0,0 +1,27 @@
...
+impl Web3Rpc {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+#[async_trait]
+impl Web3ApiServer for Web3Rpc {
+    fn client_version(&self) -> Result<String> {
+        let kakarot_client_version = format!("kakarot_{}", env!("CARGO_PKG_VERSION"));
```
bajpai244, line 19: This can also be a lazy static, for now I went with this approach, but if you think a lazy static is better, then I can do that as well.

crates/eth-rpc/src/servers/web3_rpc.rs
bajpai244: Also, do we think we can make a low priority issues for the above , if yes then I can create that.

crates/eth-rpc/src/servers/web3_rpc.rs
ClementWalter: maybe a `lazy_static` in `constants` to keep all the constant definition at the same place indeed?

crates/eth-rpc/src/servers/web3_rpc.rs
ClementWalter: I think that there has been a long confusion in the two terms and eth eventually migrated to keccak256, maybe this route is old? https://ethereum.stackexchange.com/questions/550/which-cryptographic-hash-function-does-ethereum-use

ClementWalter (commented):
nice!

crates/eth-rpc/src/servers/web3_rpc.rs
bajpai244: Make sense! Changed.

crates/eth-rpc/src/servers/web3_rpc.rs
bajpai244: Oh, so what I am talking about here is that in Ethereum, hashing is Keccak256, so for L1 execution client, this endpoint should return Keccak256.

In zk-evms, usually we do use zk-friendly hashing algorithms , from the daily we had I got to know that we use a mix of hashing algorithm.

So, I think we should investigate that the tools that use this endpoint, what do they usually use it for, and then see if there is some other hashing algorithm that we need to use here. { although this should not be a priority, as we aren't sure how much is this endpoint used. }

Lemme know if you need more clarity.

ClementWalter (approved):

PR #314, opened 2023-07-14
feat: add sharingan for starknet stack

ftupas:
Time spent on this PR:

# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

Kindly see https://github.com/keep-starknet-strange/madara-infra/issues/7 for context. This pr does the following:
- add sharingan as a network for the RPC
- update supported platforms for binaries

- [ ] Yes
- [x] No

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN rustup self update
 RUN cargo build --all --release
 
 # Create a new container from scratch to reduce image size
-FROM debian:buster-slim
+FROM debian:bullseye
 
 # Install any necessary dependencies
 RUN apt-get update && apt-get install -y libssl-dev ca-certificates tini && rm -rf /var/lib/apt/lists/*
@@ -37,4 +37,3 @@ ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/bin/kakarot-rpc"]
 # empty CMD is needed and cannot be --help because otherwise configuring from
 # environment variables only would be impossible and require a workaround.
 CMD []
-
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN rustup self update
 RUN cargo build --all --release
 
 # Create a new container from scratch to reduce image size
-FROM debian:buster-slim
+FROM debian:bullseye
```
ftupas, line 15: Had to change update this due to https://github.com/rust-lang/docker-rust/issues/77

crates/core/src/client/config.rs
```diff
@@ -36,6 +37,7 @@ impl Network {
         match self {
             Network::Katana => Ok(Url::parse(KATANA_RPC_URL)?),
             Network::Madara => Ok(Url::parse(MADARA_RPC_URL)?),
+            Network::Sharingan => Ok(Url::parse(std::env::var("SHARINGAN_RPC_URL").unwrap().as_str())?),
```
ftupas, line 40: This means `SHARINGAN_RPC_URL` needs to be passed when network used is sharingan

danilowhk, line 40: Isnt it possible to have the same format as how we access Madara and Katana Url?

ftupas: I opted to have this as an environment variable and not a constant like the other two as it can change depending on the deployment i.e. part of the madara stack docker compose (http://host.docker.internal:9944), stand-alone (https://sharingan.madara.zone or something else in the future)

danilowhk (approved):

PR #316, opened 2023-07-15
Update Kakarot

ClementWalter:
Update to latest commit using 

```
git submodule update --recursive --remote
```

tbh I don't know when this should be run, but I need this to have the nonce fix from the RPC point of view!


greged93 (approved):
lgtm

Issue #322, opened 2023-07-17
epic: Benchmark TPS of native Eth transfer on Kakarot


bajpai244:
# Feature Request

The feature request is to implement TPS benchmarks for native Eth transfer on Kakarot, this can help us get some meaningful early figures about the system.

**Describe Preferred Solution**

We can use [Artilley](https://www.artillery.io/) to explore this.

```[tasklist]
### Tasks
- [x] Run benchmarks locally on Madara
- [x] Run benchmarks on testnet
- [x] Run benchmarks in CI
```

Eikix:
Making this into a macro-task (aka epic) since it is quite a substantial issue

bajpai244:
Breaking this tasks into two PR:
- Benchmarks on Madara
- Benchmarks on Testnet

I have created two issues for it as well, the epic can be completed done when PRs for both the below issues are merged:
- Issue for benchmarks on Madara
- Issue for benchmarks on Testnet

Eikix:
Going to run our test quickly to see the result

PR #324, opened 2023-07-17
feat: add native token transfer benchmarks

bajpai244:
This PR is part of the epic #322 , and resolves the issue with #374.

The PR does the following:
- Adds a `benchmarking` directory which is self-sufficient for benchmarking Kakarot on Madara
- Adds Madara as a submodule, the build of which is used at the time of benchmarking
- Adds `benchmarking.yml` to the github workflows, so that we can do these benchmarking at github actions
    - The workflow run on each pull request and pr to main
    -  It will compare the performance of commits from new PR with the previous historical data
    - Once a PR is merged, it will automatically create a website on the branch gh:pages to host these benchmarks via GitHub pages, if it has already created this website then it will update the data for the website.
- `README.md` of `benchmarking` directory has all instructions on how to run them yourselves.

NOTE:
before merging this PR, we need a `gh-pages` repository to be present, where the action will keep pushing data and will contain website to show it on charts.

How to create this:
```
# Create a local branch
$ git checkout --orphan gh-pages

// add readme
$ git add README.md

// commit
$ git commit

// This has to be pushed to kakarot-rpc and not your local fork
$ git push upstream gh-pages
```

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,106 @@
+---
+    name: Task - Benchmark
+    
+    on:
+      push:
+        branches: [main]
+      pull_request:
+        branches: [main]
+
+    permissions:
+      deployments: write
+      contents: write
+      pull-requests: write
+    
+    jobs:
+      benchmark:
+        runs-on: ubuntu-latest
+        steps:
+          - uses: actions/checkout@v3
+            with:
+              clean: false
+          
+          # install dependencies
+          - name: Install submodules and dependencies
+            run: |
+              git submodule update --init --recursive
+              sudo apt update
+              sudo apt install -y protobuf-compiler clang jq
+
+          - uses: actions/cache@v3
+            id: cache
+            with:
+              path: |
+                ~/.cargo/bin/
+                ~/.cargo/registry/index/
+                ~/.cargo/registry/cache/
+                ~/.cargo/git/db/
+                ./lib/madara/target/
+              key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{hashFiles('./lib/madara/Cargo.lock')}}
+          
+          # build rpc and madara  
+          - name: Setup Rust
+            uses: actions-rs/toolchain@v1
+            with:
+                toolchain: stable
+          - name: Build RPC
+            run: |
+              cargo build --release
+          - name: Build Madara
+            if: steps.cache.outputs.cache-hit  != 'true'
+            run: |
+              cd ./lib/madara
+              cargo build --release
+          
+          
+          # Python Setup and dependencies installation
+          - uses: actions/setup-python@v4
+            with:
+              python-version: "3.9.13"
+          - name: Load cached Poetry installation
+            id: cached-poetry
+            uses: actions/cache@v3
+            with:
+              path: ~/.local
+              key: poetry-${{ runner.os }}
+          - name: Install Poetry
+            if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+            uses: snok/install-poetry@v1
+            with:
+              virtualenvs-create: true
+              virtualenvs-in-project: true
+              installer-parallel: true
+          - name: install kakarot python dependencies
+            run: |-
+              cd lib/kakarot
+              poetry install 
+          - name: Load cached venv
+            id: cached-poetry-dependencies
+            uses: actions/cache@v3
+            with:
+              path: .venv
+              key: venv-${{ runner.os }}-${{ hashFiles('./lib/kakarot/poetry.lock') }}
+
+          # benchmark rpc
+          - name: install benchmark projects
+            run: |-
+              cd ./benchmarking
+              npm i
+          - name: Run benchmark
+            run: |
+              cd benchmarking
+              ./scripts/benchmark.sh
+
+          # compare benchmarks and publish
+          - name: Compare result
+            uses: benchmark-action/github-action-benchmark@v1
+            with:
+              tool: "customBiggerIsBetter"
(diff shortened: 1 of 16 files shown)


benchmarking/prettier.json
ClementWalter, line 1: there is already a `.prettierrc.json` at the root of the project, can they be merged?

benchmarking/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "name": "kakarot-rpc-benchmarking",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "type": "commonjs",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "format": "npx prettier --write .",
+    "metrics": "node ./scripts/metrics.js",
+    "run-madara": "RUST_LOG=error ../lib/madara/target/release/madara --dev",
```
ClementWalter, line 11: when is madara built actually?

benchmarking/README.md
```diff
@@ -0,0 +1,31 @@
+# Kakarot Benchmarking
+
+The directory contains scripts which we use to benchmark Kakarot, as of now we
+are using aritillery to create virtual users and test the RPC.
+
+NOTE: If you have a `.env` file at the root of this repository, then it is suggested to rename it to something else when you run these benchmarks, the reason being that the RPC will always override variables from `.env`, which can lead to benchmarks not working.
```
ClementWalter, line 6: lint your file with prettier!

benchmarking/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Steps to benchmark:
+
+- install all submodules in the project root:
+  - git submodule update --init --recursive
+- install Kakarot dependencies and compile:
+  - from the project root:
+    - cd lib/kakarot
+    - poetry install { creating a venv is advised }
+    - STARKNET_NETWORK=madara poetry run python ./scripts/compile_kakarot.py
```
ClementWalter, line 16: we'd need to deploy as well I guess. But I'm wondering if we could just use `docker compose up` for all the setting

.github/workflows/benchmark.yml
ClementWalter, line 1: why do you have a left spacing for the whole file?

ClementWalter (approved):
I'd go with a merge if it works but overall the setup seems too complicated to me. I think that we should use a compose to have everything up and running and then pure ethers.js/artillery scripts targeting the RPC endpoint

.github/workflows/benchmark.yml
bajpai244: On my end I cannot see any spacing.

benchmarking/README.md
bajpai244: The reason for not using docker was to get as close to real TPS as possible. I know docker doesn't emulate, but I just wanted to avoid any overhead that could be caused by it.

We can later on experiment with a setup via docker and if we see similar performance then we can shift!

Lemme know what you think about it.

benchmarking/prettier.json
bajpai244: I have merged them.

benchmarking/README.md
bajpai244: done.

benchmarking/package.json
bajpai244: In the CI, `benchmark.yml`.

image

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,107 @@
...
+              tool: "customBiggerIsBetter"
+              output-file-path: ./benchmarking/reports/metrics.json
+              alert-threshold: "120%"
+              github-token: ${{ secrets.GITHUB_TOKEN }}
+              fail-on-alert: true
+              summary-always: ${{ github.ref == 'refs/heads/main' }}
+              comment-always: false
+              comment-on-alert: false
+              auto-push: ${{ github.ref == 'refs/heads/main' }}
+
```
ftupas, line 107: nit: newline

.prettierignore
```diff
@@ -0,0 +1,2 @@
+lib
+target
```
ftupas, line 2: newline here as well

benchmarking/.gitignore
```diff
@@ -0,0 +1,6 @@
+# node modules
+node_modules
+
+# reports dir, keeps benchmark reports
+reports/*
+!reports/.gitkeep
```
ftupas, line 6: newline here as well

benchmarking/scripts/benchmark.sh
```diff
@@ -0,0 +1,2 @@
+npm run benchmark:ci
+exit 0
```
ftupas, line 2: newline here as well

benchmarking/scripts/deploy_kakarot.sh
```diff
@@ -0,0 +1,41 @@
...
+    exit 1
+    fi
+
+cd ..
+
+# make pull-kakarot
+cp lib/kakarot/.env.example lib/kakarot/.env
+cd lib/kakarot 
+STARKNET_NETWORK=madara poetry run python ./scripts/compile_kakarot.py
+STARKNET_NETWORK=madara poetry run python ./scripts/deploy_kakarot.py
```
ftupas, line 41: I think use the make command `cd lib/kakarot && STARKNET_NETWORK=madara make deploy`

ftupas (commented):
Looks good! Just a smol comment and some nits

benchmarking/scripts/deploy_kakarot.sh
bajpai244: done.

ftupas (approved):
Looks good to me! 🚀

Issue #310, opened 2023-07-13
dev: use namespaces for defining API traits


bajpai244:
Current we are adding prefix { like eth_ , alchemy_ } to all the jsonrpsee APIs we are defining, we can use [namespaces](https://github.com/paritytech/jsonrpsee/blob/3628ba161c42e9677d506ba94734e4d6de968901/proc-macros/src/lib.rs#L81) for that.
  
**Describe Preferred Solution**

Replace all jsonrpsee api definitions to use namespace.

```[tasklist]
### Tasks
- [ ] Refactor all APIs defined [here](https://github.com/kkrt-labs/kakarot-rpc/tree/main/crates/eth-rpc/src/api) here to use namespaces, and remove the corresponding prefixes.
- [ ] Refactor `Net` method from Eth Api to seperate Net API
```

PR #329, opened 2023-07-20
refactor: use namespaces to remove using prefixes for annotating methods, refactor Net API api …

bajpai244:
Refactor which allows us to use namespaces for all APIs, which removes the need for adding prefixes like { eth_, alchemy_ } while annotating these methods.

Also, this PR refactors the Net Api method from our from Eth Api, and defines and implements it as a seperate API { Net is a seperate API, and reth treats it like it as well, you can check it out [here](https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-api/src/net.rs). }

Resolves: #310 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

- use of namespaces in APIs.
- net is refactored and implemented as a separate API.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/eth-rpc/src/api/eth_api.rs
+++ b/crates/eth-rpc/src/api/eth_api.rs
@@ -7,10 +7,10 @@ use reth_rpc_types::{
     TransactionReceipt, TransactionRequest, Work,
 };
 
-#[rpc(server)]
+#[rpc(server, namespace = "eth")]
 #[async_trait]
 pub trait EthApi {
-    #[method(name = "eth_blockNumber")]
+    #[method(name = "blockNumber")]
     async fn block_number(&self) -> Result<U64>;
 
     /// Returns the protocol version encoded as a string.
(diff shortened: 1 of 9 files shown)


bajpai244:
I have also created a low priority issue to define and implement all methods of the Net API [here](#330 ).

crates/eth-rpc/src/api/eth_api.rs
```diff
@@ -7,106 +7,106 @@ use reth_rpc_types::{
...
-#[rpc(server)]
+#[rpc(server, namespace = "eth")]
 #[async_trait]
 pub trait EthApi {
-    #[method(name = "eth_blockNumber")]
+    #[method(name = "blockNumber")]
     async fn block_number(&self) -> Result<U64>;
 
     /// Returns the protocol version encoded as a string.
     #[method(name = "net_version")]
```
ClementWalter, line 17: what about this `net_` method? will it become `eth_net_version` ? now that we've split things in namespaces, we may just add a `net` namespace?

ClementWalter: just saw that it's what you did! but then we need to remove this right?

crates/eth-rpc/src/api/net_api.rs
```diff
@@ -0,0 +1,12 @@
+use jsonrpsee::core::RpcResult as Result;
+use jsonrpsee::proc_macros::rpc;
+use reth_primitives::U64;
+
+// TODO: Define and implement of methods of Net API
```
ClementWalter, line 5: what are the other API for this namespace? could we possibly add `todo!()` and then create issues, and then let contrib work on them ?

crates/eth-rpc/src/api/net_api.rs
```diff
@@ -0,0 +1,12 @@
...
+#[rpc(server, namespace = "net")]
+#[async_trait]
+pub trait NetApi {
+    /// Returns the protocol version encoded as a string.
+    #[method(name = "version")]
+    fn protocol_version(&self) -> Result<U64>;
```
ClementWalter, line 11: for information, but why calling this with the `protocol_` prefix?

ClementWalter (changes requested):
Need to remove `net_version` from the `eth` namespace

crates/eth-rpc/src/api/eth_api.rs
bajpai244: oh! I thought I have removed it from here as well! Thanks for the catch.

crates/eth-rpc/src/api/net_api.rs
bajpai244: make sense, I have added todo!().

crates/eth-rpc/src/api/net_api.rs
bajpai244: This is how it [earlier](https://github.com/kkrt-labs/kakarot-rpc/blob/165d27a87fbd54a8430399c76af171c67511bb9f/crates/eth-rpc/src/servers/eth_rpc.rs#L41) was, but I have renamed it to version.

crates/eth-rpc/src/api/net_api.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn version(&self) -> Result<U64>;
+
+    /// Returns number of peers connected to node.
+    #[method(name = "peerCount")]
+    fn peer_count(&self) -> Result<PeerCount>;
+
+    /// Returns true if client is actively listening for network connections.
+    /// Otherwise false.
+    #[method(name = "listening")]
+    fn is_listening(&self) -> Result<bool>;
```
ClementWalter, line 21: why is this `is_listening` and not `listening`?

ClementWalter (commented):
just a minor question regarding `is_listening!`!

crates/eth-rpc/src/api/net_api.rs
bajpai244: I took the function name from my investigation into [reth](https://github.com/paradigmxyz/reth/blob/15781beda869c5db682bb408d8228b90045d3471/crates/rpc/rpc-api/src/net.rs#L19), the name made sense to me as its return type is a boolean.

But, I think `listening` makes sense as well, I have refactored.

bajpai244:
> just a minor question regarding `is_listening!`!

resolved.

ClementWalter (approved):

Issue #216, opened 2023-06-28
fix: eth_call Erroring when being called by parent contract


danilowhk:
# Bug Report

Currently eth_call endpoint is returning an error when being called by another from another contract

**Kakarot RPC version:**

**Current behavior:**

**Expected behavior:**

**Steps to reproduce:**

**Related code:**

Contract being tested

```
// SPDX-License-Identifier: MIT

pragma solidity ^0.8.6;

contract Counter {
    uint256 private count = 0;

    function increment() public {
        count += 1;
    }

    function getCount() public view returns (uint256) {
        return count;
    }
}

contract Caller {
    Counter private counter;

    constructor(address _counterAddress) {
        counter = Counter(_counterAddress);
    }

    function callIncrement() public {
        counter.increment();
    }

    function getCounter() public view returns (uint256) {
        return counter.getCount();
    }
}

```

**Other information:**

Eikix:
What is the status of this @danilowhk ? 
Is this directly linked with https://github.com/kkrt-labs/kakarot/issues/617 ?

Where in the kakarot main SC code can we see that subcalls use get_call_address()?

danilowhk:
Doesnt seem related as the error messages are different.

And from what was tested with @ClementWalter , it seems the issue is on the RPC as when trying to do subcalls directly from python script, the tests passed (even against Madara)

Eikix:
> Doesnt seem related as the error messages are different.
> 
> And from what was tested with @ClementWalter , it seems the issue is on the RPC as when trying to do subcalls directly from python script, the tests passed (even against Madara)


When you say subcalls directly from python scripts, do you mean subcalls, in the context of a "view" call (not invoke) cross contract, on Madara?

danilowhk:
It seems the script is calling as "view":

https://github.com/kkrt-labs/kakarot/blob/baa08057bf24c533748883fe2551a148d6cfcba3/scripts/utils/kakarot.py#L128

This is the code for the script:

```import os

os.environ["STARKNET_NETWORK"] = "madara"
from scripts.utils.kakarot import *

counter = await deploy("PlainOpcodes", "Counter")
plain_opcodes = await deploy("PlainOpcodes", "PlainOpcodes", counter.address)


assert await plain_opcodes.opcodeStaticCall() == await counter.count()
await plain_opcodes.opcodeCall()
await counter.count()
await plain_opcodes.opcodeStaticCall()```

[...]

PR #359, opened 2023-07-27
demonstrate kkrt handling the `get_caller_address` zero case w/ error

jobez:
Time spent on this PR: 2 hrs

Resolves: #

addresses: #216 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [X] Testing

# What is the new behavior?

This small pr introduces a potential pattern where pre-alpha testers can do PR's w/ solidity contracts. Currently using testsequencer, which is a little involved, the idea is to have a `solidity_contracts` directory that pre-alpha tests can add solidity code that isnt working and have those contracts ran against the CI

- [ ] Yes
- [X] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,16 @@ pull-kakarot: .gitmodules
 build-kakarot: setup 
 	cd lib/kakarot && make build && make build-sol
 
+build-testing-solidity-contracts:
+	FOUNDRY_PROFILE=testing forge build
+
 build-and-deploy-kakarot:
 	cd lib/kakarot && STARKNET_NETWORK=$(STARKNET_NETWORK) make deploy
 
 deploy-kakarot:
 	cd lib/kakarot && STARKNET_NETWORK=$(STARKNET_NETWORK) poetry run python ./scripts/deploy_kakarot.py
 
-setup: pull-kakarot build-kakarot
+setup: pull-kakarot build-kakarot build-testing-solidity-contracts
 
 # run devnet
 devnet: 
--- a/crates/core/tests/client.rs
+++ b/crates/core/tests/client.rs
@@ -129,6 +129,52 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_get_caller_address() {
+        let starknet_test_sequencer = construct_kakarot_test_sequencer().await;
+
+        let expected_funded_amount = FieldElement::from_dec_str("1000000000000000000").unwrap();
+
+        let deployed_kakarot =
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -12,13 +12,16 @@ pull-kakarot: .gitmodules
 build-kakarot: setup 
 	cd lib/kakarot && make build && make build-sol
 
+build-testing-solidity-contracts:
+	FOUNDRY_PROFILE=testing forge build
```
jobez, line 16: testing profile allows us to specify two different source directories. i am leaving the `out` compilation output as the same, so i wouldnt need to dig too deeply in the assumptions that the `deploy_helpers` code I wrote makes about where to look for deployed contracts

crates/core/tests/client.rs
```diff
@@ -128,6 +128,52 @@ mod tests {
         assert_eq!(num, 1);
     }
 
+    #[tokio::test]
+    async fn test_get_caller_address() {
```
jobez, line 132: currently calling the testing solidity with the testsequencer. ideal flow would be to have an input solidity devs are comfortable with that is ran against our ci.

jobez (commented):
brief description of justification for the changes

solidity_contracts/GetCallerAddress.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+contract GetCallerAddress {
+    // This function will return the address of the caller
+    function getCallerAddress() public view returns (address) {
+        return msg.sender;
```
Eikix, line 7: this should return 0 or any caller address, right? we're just checking that it doesnt error out

> It appears you're running a call, which runs quickly and does not alter the state of the blockchain. msg.sender is set in both a transaction and a call. In a transaction, it cannot be faked: you must have the private key associated with the given account. But in a call, you are free to set the sender to any value you like.

from:
https://stackoverflow.com/questions/51847788/msg-sender-does-not-work-inside-a-view-function-why-is-there-a-workaround

ClementWalter:
As discussed this morning, I don't feel like we need to add this kind of tests here. For the RPC I'd like us to focus first on the conformance tests.

Regarding this `CALLER` (and `ORIGIN`) opcode issue, we made some tests with @danilowhk to make sure that we understood correctly the expected behavior of the EVM, and tl;dr added an issue for adding `from` and `origin` to the `ExecutionContext`.

I'd like to close this issue here and focus on this task on the main kakarot repo side.

Issue #368, opened 2023-07-28
bug: tests failing on main due to private function import


ftupas:
# Bug Report

**Kakarot RPC version:**

**Current behavior:**

<img width="917" alt="image" src="https://github.com/kkrt-labs/kakarot-rpc/assets/35031356/a5382c92-325f-4be7-add5-e24a7a0e54c8">

**Expected behavior:**

Test should pass

**Steps to reproduce:**

**Related code:**

```bash
cargo test
```

**Other information:**

PR #369, opened 2023-07-28
fix: fix import error in balance tests

Eikix:
bugfix


ftupas (approved):
Looks good ⚡️ thanks for the blazing fast fix

Eikix:
solves #368

Issue #365, opened 2023-07-27
dev: add integration test environment for `eth-rpc`


ftupas:
**Describe the enhancement request**

It would be great to be able to do the following
- launch test sequencer
- deploy kakarot system
- launch kakarot client
- launch eth-rpc with kakarot client
- do the test against test sequencer

**Describe Preferred Solution**

Build off of @jobez's implementation https://github.com/kkrt-labs/kakarot-rpc/pull/284/files

PR #370, opened 2023-07-28
dev: add integration test environment for `eth-rpc`

ftupas:
Time spent on this PR: 0.5

Resolves: #365 

# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

- Add integration test environment for Kakarot RPC with the following
  - TestSequencer
  - Deployed Kakarot contracts
  - Kakarot RPC server
- Move `test_utils` so it can be used in other crates

- [ ] Yes
- [X] No

### DIFF
--- a/crates/eth-rpc/src/test_utils.rs
+++ b/crates/eth-rpc/src/test_utils.rs
@@ -1,0 +1,90 @@
+use std::net::SocketAddr;
+use std::sync::Arc;
+
+use dojo_test_utils::sequencer::TestSequencer;
+use jsonrpsee::server::ServerHandle;
+use kakarot_rpc_core::client::config::{Network, StarknetConfig};
+use kakarot_rpc_core::client::KakarotClient;
+use kakarot_rpc_core::test_utils::constants::EOA_WALLET;
+use kakarot_rpc_core::test_utils::deploy_helpers::deploy_kakarot_system;
+use starknet::core::types::FieldElement;
+use starknet::providers::jsonrpc::HttpTransport as StarknetHttpTransport;
+use starknet::providers::JsonRpcClient as StarknetJsonRpcClient;
+
+use crate::config::RPCConfig;
+use crate::rpc::KakarotRpcModuleBuilder;
+use crate::run_server;
+
+/// Sets up the environment for Kakarot RPC integration tests by deploying the Kakarot contracts
+/// and starting the Kakarot RPC server.
+///
+/// This function:
+/// 1. Takes an `Arc<TestSequencer>` as input, which is used to deploy the Kakarot contracts and to
+///    set up the Kakarot RPC server.
+/// 2. Deploys the Kakarot contracts.
+/// 3. Creates Starknet and Kakarot clients.
+/// 4. Sets up and runs the Kakarot RPC module.
+///
+/// # Arguments
+///
+/// * `starknet_test_sequencer` - An Arc-wrapped TestSequencer. This is used to deploy the Kakarot
+///   contracts and to set up the Kakarot RPC server.
+///
+/// # Returns
+///
+/// This function returns a Result containing a tuple with the server's address and a handle to
+/// stop the server upon successful execution.
+///
+/// The function may return an Err variant of eyre::Report if there are issues with deploying the
+/// Kakarot contracts, creating the clients, or running the RPC server.
+///
+/// # Example
+/// ```ignore
+/// use kakarot_rpc::test_utils::setup_kakarot_rpc_integration_env;
+/// use dojo_test_utils::sequencer::TestSequencer;
+/// use std::sync::Arc;
+/// use tokio::runtime::Runtime;
+///
+/// #[tokio::test]
+/// async fn test_case() {
+///    // Create a TestSequencer.
+///    let test_sequencer = Arc::new(TestSequencer::new());
+///
+///    // Set up the Kakarot RPC integration environment.
+///    let (server_addr, server_handle) = setup_kakarot_rpc_integration_env(&test_sequencer).await.unwrap();
+///    
+///    // Query whatever eth_rpc endpoints
+///     
+///    // Dont forget to close server at the end.
+///    server_handle.stop().expect("Failed to stop the server");
+///
+/// }
+/// ```
+pub async fn setup_kakarot_rpc_integration_env(
+    starknet_test_sequencer: &Arc<TestSequencer>,
+) -> Result<(SocketAddr, ServerHandle), eyre::Report> {
+    // Define the funding amount.
+    let funding_amount = FieldElement::from_dec_str("1000000000000000000")?;
+
+    // Deploy Kakarot contracts.
+    let deployed_kakarot = deploy_kakarot_system(starknet_test_sequencer, EOA_WALLET.clone(), funding_amount).await;
+
+    // Initialize StarknetHttpTransport with sequencer's URL.
+    let starknet_http_transport = StarknetHttpTransport::new(starknet_test_sequencer.url());
+
+    // Create Starknet and Kakarot clients.
+    let starknet_config = StarknetConfig::new(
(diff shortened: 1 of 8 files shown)


crates/eth-rpc/tests/utils.rs
```diff
@@ -35,3 +47,77 @@ pub async fn setup_mock_eth_rpc() -> KakarotEthRpc<JsonRpcClient<MockJsonRpcTran
...
+/// # Returns
+///
+/// This function returns a Result containing a tuple with the server's address and a handle to
+/// stop the server upon successful execution.
+///
+/// The function may return an Err variant of eyre::Report if there are issues with deploying the
+/// Kakarot contracts, creating the clients, or running the RPC server.
+///
+/// # Example
+/// ```ignore
```
Eikix, line 75: I think this ``` is not closed

Eikix (commented):
Very cool PR! A minor comment but lgtm, letting @ClementWalter or @jobez take a look

crates/eth-rpc/tests/utils.rs
```diff
@@ -1,11 +1,23 @@
...
 use kakarot_rpc_core::mock::mock_starknet::{all_fixtures, init_mock_client};
-use starknet::providers::JsonRpcClient;
+use kakarot_rpc_core::test_utils::constants::EOA_WALLET;
+use kakarot_rpc_core::test_utils::deploy_helpers::deploy_kakarot_system;
+use starknet::core::types::FieldElement;
+use starknet::providers::jsonrpc::HttpTransport as StarknetHttpTransport;
+use starknet::providers::{JsonRpcClient, JsonRpcClient as StarknetJsonRpcClient};
 
-/// Run wiremock to fake starknet rpc and then run our own `kakarot_rpc_server`.
+/// Run wiremock to fake Starknet rpc and then run our own `kakarot_rpc_server`.
```
ClementWalter, line 20: I thought that we removed `wiremock`, so probably needs an update

crates/eth-rpc/tests/utils.rs
```diff
@@ -35,3 +47,78 @@ pub async fn setup_mock_eth_rpc() -> KakarotEthRpc<JsonRpcClient<MockJsonRpcTran
...
+///    // Create a TestSequencer.
+///    let test_sequencer = Arc::new(TestSequencer::new());
+///
+///    // Set up the Kakarot RPC integration environment.
+///    let (server_addr, server_handle) = setup_kakarot_rpc_integration_env(&test_sequencer).await.unwrap();
+///    
+///    // Query whatever eth_rpc endpoints
+///     
+///    // Dont forget to close server at the end.
+///    server_handle.stop().expect("Failed to stop the server");
```
ClementWalter, line 92: can't we have this automatically done in `Drop` for example?

crates/eth-rpc/tests/utils.rs
```diff
@@ -35,3 +47,78 @@ pub async fn setup_mock_eth_rpc() -> KakarotEthRpc<JsonRpcClient<MockJsonRpcTran
...
+///
+/// }
+/// ```
+pub async fn setup_kakarot_rpc_integration_env(
+    starknet_test_sequencer: &Arc<TestSequencer>,
+) -> Result<(SocketAddr, ServerHandle), eyre::Report> {
+    // Define expected funded amount.
```
ClementWalter, line 99: why is this "expected"?

crates/eth-rpc/tests/rpc.rs
```diff
@@ -1,5 +1,5 @@
 #![recursion_limit = "1024"]
-mod utils;
+pub mod utils;
```
ClementWalter, line 2: why do we need to `pub` this now?

crates/eth-rpc/tests/integration.rs
```diff
@@ -0,0 +1,40 @@
+pub mod utils;
```
ClementWalter, line 1: why `pub` ?

crates/eth-rpc/tests/integration.rs
```diff
@@ -0,0 +1,40 @@
...
+    #[tokio::test]
+    async fn test_get_block_rpc() {
+        // Setup the test sequencer
+        let starknet_test_sequencer = Arc::new(construct_kakarot_test_sequencer().await);
+
+        // Deploy the Kakarot contracts and start the Kakarot RPC server
+        let (server_addr, server_handle) = setup_kakarot_rpc_integration_env(&starknet_test_sequencer)
+            .await
+            .map_err(|e| {
+                println!("Error setting up Kakarot RPC server: {}", e);
```
ClementWalter, line 23: not sure to understand why we don this `map_err`; is this to sort of do a manual logging? ain't there any cleaner way to do so?

crates/eth-rpc/tests/integration.rs
```diff
@@ -0,0 +1,40 @@
...
+                e
+            })
+            .unwrap();
+
+        // Try to run the test
```
ClementWalter, line 28: Do we "try to run" the test, or do we actually run it? ;)

crates/eth-rpc/tests/integration.rs
```diff
@@ -0,0 +1,40 @@
...
+        let provider = EthersHttp::from_str(format!("http://localhost:{}", server_addr.port()).as_ref()).unwrap();
+        let block_number: U64 =
+            ethers::prelude::JsonRpcClient::request(&provider, "eth_blockNumber", ()).await.unwrap();
+        let params = (block_number, true);
+        let block: std::result::Result<EthersBlock<EthersH256>, _> =
+            ethers::prelude::JsonRpcClient::request(&provider, "eth_getBlockByNumber", params).await;
```
ClementWalter, line 34: not sure about what we want to test here? why do we call these two routes especially? I guess that this is in preparation for hive tests right? but this test confuses we

crates/eth-rpc/tests/integration.rs
ftupas: Indeed we actually run it! haha 😅

crates/eth-rpc/tests/integration.rs
ftupas: I wanted to test that by calling `eth_getBlockByNumber` endpoint, we are getting a block given all systems being deployed, so for that I'm passing the `block_number` as parameter which I get from the `eth_blockNumber` endpoint

crates/eth-rpc/tests/utils.rs
ClementWalter: mock yes, `wiremock` no

crates/eth-rpc/tests/integration.rs
ClementWalter: but in the `rpc.rs` case for example, it was not `pub` before and it was working, and you added `pub` as well. I still don't see why we would need to add this. We may need to, but I'd like to understand first

ClementWalter (commented):
we need to understand this `pub` issue

crates/eth-rpc/src/test_utils.rs
```diff
@@ -0,0 +1,90 @@
...
+///    let (server_addr, server_handle) = setup_kakarot_rpc_integration_env(&test_sequencer).await.unwrap();
+///    
+///    // Query whatever eth_rpc endpoints
+///     
+///    // Dont forget to close server at the end.
+///    server_handle.stop().expect("Failed to stop the server");
+///
+/// }
+/// ```
+pub async fn setup_kakarot_rpc_integration_env(
```
ftupas, line 63: Finally moved this to `test_utils` so it can be used in integration tests i.e. conformance test

ftupas: No more `pub` issue on the `utils.rs` file

ClementWalter (approved):

PR #389, opened 2023-08-02
feat: modify genesis set storage contract account

ftupas:
Time spent on this PR: 0.5

Resolves: #379 

# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

- We are now able to set the storage of a Kakarot contract account
- Refactor code to it more DRY
- This is a continuation of #383 

- [ ] Yes
- [X] No

### DIFF
--- a/crates/conformance-test-utils/src/madara.rs
+++ b/crates/conformance-test-utils/src/madara.rs
@@ -4,11 +4,11 @@ use reth_primitives::{Bytes, U128, U256};
 use starknet::core::types::FieldElement;
 use starknet::core::utils::get_storage_var_address;
 
-use crate::types::{ContractAddress, Felt, StorageKey, StorageValue};
+use crate::types::{ContractAddress, StorageKey, StorageValue};
 
 pub fn genesis_load_bytecode(
     bytecode: &Bytes,
-    address: FieldElement,
+    starknet_address: FieldElement,
 ) -> Vec<((ContractAddress, StorageKey), StorageValue)> {
     bytecode
         .chunks(16)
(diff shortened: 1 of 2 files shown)


crates/conformance-test-utils/src/madara.rs
```diff
@@ -272,4 +323,117 @@ mod tests {
...
+
+        // When
+        let result = genesis_set_storage_kakarot_contract_account(starknet_address, key, value).unwrap();
+        // Then
+        assert_eq!(result, expected_output);
+    }
+
+    #[tokio::test]
+    async fn test_kakarot_contract_account_storage() {
+        let starknet_test_sequencer = construct_kakarot_test_sequencer().await;
```
ClementWalter, line 372: when https://github.com/kkrt-labs/kakarot-rpc/pull/390 is merged, you'll be able to rebase and simplify this

ftupas: Yeah we can remove a lot of this boilerplate code 💪

crates/conformance-test-utils/src/madara.rs
```diff
@@ -272,4 +323,117 @@ mod tests {
...
+        // Assert that the value stored in the contract is the same as the value we set in the genesis
+        assert_eq!(expected_value, actual_value);
+    }
+
+    #[test]
+    fn test_split_u256_into_field_elements() {
+        // Given
+        let value = U256::MAX;
+        let expected_output = [
+            FieldElement::from_hex_be("0xffffffffffffffffffffffffffffffff").unwrap(),
```
ClementWalter, line 429: plz make the test more sound, ofc for this we'd probably want to have a fuzzer, but we can at least make sure that high * 2**128 + low = input, and use an input such that high != low

ftupas: Gotcha, I've update this test to recalculate the `U256` from the resulting `FieldElement` values and assert equality. provided some test cases 🤙

crates/conformance-test-utils/src/madara.rs
```diff
@@ -274,4 +326,130 @@ mod tests {
...
+        let actual_value =
+            genesis_contract.storage(&key_low, &key_high, &StarknetBlockId::Tag(BlockTag::Latest)).await.unwrap();
+
+        // Assert that the value stored in the contract is the same as the value we set in the genesis
+        assert_eq!(expected_value, actual_value);
+    }
+
+    #[test]
+    fn test_split_u256_into_field_elements() {
+        let test_cases = vec![
```
ClementWalter, line 430: this is ok for this PR but generally speaking you **don't** want to loop over test cases **inside** the test but rather to _parametrize_ the test because you indeed have 3 tests here and not one: you don't want to have all the tests failing when one case is failing. In python, we used `@pytest.mark.parametrize`, in rust it seems like we could try to use https://github.com/la10736/rstest, would you be interested in trying to add this to the project and do some test refactoring?

ftupas: Nice was actually thinking to explore some how to have fixtures and parametrize tests in Rust, I'd be keen to include it in the project! Let me create an issue 💪

ClementWalter (approved):
I created an issue for refactoring the for loop inside your test https://github.com/kkrt-labs/kakarot-rpc/issues/393

PR #432, opened 2023-08-16
fix: add dependencies as skip-label to not fail dependabot CI job

Eikix:
smol pr to solve cumbersome adding label by hand


ftupas (approved):
Nice, looks good to me!

Issue #436, opened 2023-08-16
feat: add deployer account and deploy fee to Madara genesis config


ftupas:
# Feature Request

**Describe the Feature Request**

With this [pr](https://github.com/kkrt-labs/kakarot/pull/674/files#diff-52a0f8d2aed2b19fc2c6b24578a80469c17654a862db340d093cff7ec44e2ada), we will need to add the deployer account as one of the deployed contracts as well as set the [storage](https://github.com/kkrt-labs/kakarot/blob/8b25137e6e9937510c1642665cec4f156f6a49dd/src/kakarot/constants.cairo#L26) for Kakarot contract

**Describe Preferred Solution**

Add them in `contracts` and `storage` respectively in [`genesis.rs`](https://github.com/kkrt-labs/kakarot-rpc/blob/5ae048d81f61780bbbb811177956ba68da40465f/crates/conformance-test-utils/src/hive/genesis.rs)

PR #438, opened 2023-08-16
feat: add deployer account and deploy fee to madara genesis

ftupas:
Time spent on this PR: 0.1

Resolves: #436 

# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

Deployer account and deploy fee added to madara genesis

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No


crates/hive-utils/src/hive/genesis.rs
```diff
@@ -43,6 +43,7 @@ impl HiveGenesisConfig {
 lazy_static! {
     pub static ref KAKAROT_ADDRESSES: FieldElement = FieldElement::from_hex_be("0x9001").unwrap(); // Safe unwrap, 0x9001
     pub static ref BLOCKHASH_REGISTRY_ADDRESS: FieldElement = FieldElement::from_hex_be("0x9002").unwrap(); // Safe unwrap, 0x9002
+    pub static ref DEPLOYER_ACCOUNT_ADDRESS: FieldElement = FieldElement::from_hex_be("0x9003").unwrap(); // Safe unwrap, 0x9003
```
Eikix, line 46: can this change in another place and create a desync? or is it fixed forever

crates/hive-utils/src/hive/genesis.rs
```diff
@@ -107,6 +112,8 @@ pub async fn serialize_hive_to_madara_genesis_config(
         ("externally_owned_account", eoa_class_hash),
         ("account_proxy_class_hash", account_proxy_class_hash),
         ("blockhash_registry_address", *BLOCKHASH_REGISTRY_ADDRESS),
+        // TODO: Use DEPLOY_FEE constant https://github.com/kkrt-labs/kakarot-rpc/pull/431/files#diff-88f745498d0aaf0b185085d99a74f0feaf253f047babc85770847931e7f726c3R125
+        ("deploy_fee", FieldElement::from(100000_u64)),
```
Eikix, line 116: @bajpai244 will this become a problem? since it's not supposed to be a constant

Eikix (commented):
lgtm

crates/hive-utils/src/hive/genesis.rs
ftupas: If we initialize Madara with this config, these contracts will be fixed to these addresses

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["crates/eth-rpc", "crates/core", "crates/conformance-test-utils"]
+members = ["crates/eth-rpc", "crates/core", "crates/hive-utils"]
```
ftupas, line 2: As discussed with @greged93 , renaming this crate so it's more descriptive

crates/hive-utils/src/hive/genesis.rs
Eikix: alright

crates/hive-utils/src/hive/genesis.rs
bajpai244: So, this value can always be updated later on, u have to firstly initialize Kakarot with one `deploy_fee` which seems to be the case here, and then an off-chain service can keep updating this value, so technically we are not making this a constant here, we are just initializing Kakarot.

Eikix (approved):
lgtm

PR #461, opened 2023-08-18
dev: slightly decrease the rate at which we send transaction

bajpai244:
In the current benchmarking CI, ofthen there occurs a nonce race, and from my experience, the way to solve it is to decrease the rate at which send the transaction, this PR slightly bumps the delay between two transactions by 150ms, which should probably resolve it, otherwise we would have to move to a custom runner.

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [x] Build-related changes
- [ ] Testing

# What is the new behavior?

- Decreases the rate at which we send transactions which should lower the chance of getting a nonce race.

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### DIFF
--- a/benchmarking/scripts/processor.js
+++ b/benchmarking/scripts/processor.js
@@ -37,7 +37,7 @@ const nativeTokenTransfer = async (context, ee, next) => {
 
     // break to make sure transactions arrive in order to RPC.
     // NOTE: reduce it if you are skipping validate
-    await sleep(350);
+    await sleep(500);
   }
 
   next();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- ci: decrease the rate which we send transactions in benchmark CI
 - ci: add `CHANGELOG.md` and enforce it is edited for each PR on `main`
 - test: update integration tests to use prepopulated Katana dumped state
 - ci: cross-compile binaries to improve build time


.github/workflows/benchmark.yml
```diff
@@ -3,7 +3,7 @@
     
     on:
       push:
-        branches: [main]
+        branches: [main, dev/update-benchmark-script]
```
Eikix, line 6: is this still relevant

benchmarking/scripts/processor.js
```diff
@@ -37,7 +37,7 @@ const nativeTokenTransfer = async (context, ee, next) => {
 
     // break to make sure transactions arrive in order to RPC.
     // NOTE: reduce it if you are skipping validate
-    await sleep(350);
+    await sleep(500);
```
Eikix, line 40: why 500? could this slow tps if 500>time to post tx

Eikix (commented):
questions:) but lgtm

.github/workflows/benchmark.yml
bajpai244: nope, just added it and have removed it as well, was for local fork testing!

benchmarking/scripts/processor.js
bajpai244: So madara block time is 6second.

So, if means atleast with this we will send 12 transactions per block. So, if our TPS is less than let's say even 3, that means the pool has enough transactions, and they are not being processed!

Also, the script waits for some time for pool to be filled with transactions before collection any data.

bajpai244:
> questions:) but lgtm

have replied :)

Eikix (dismissed):
lgtm

Eikix (approved):
lgtm

PR #471, opened 2023-08-23
Dev/impl netapi methods

aniketpr01:
Time spent on this PR: 1

Resolves: https://github.com/kkrt-labs/kakarot-rpc/issues/330

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

This will return false when prompted to get values for NetApi methods as there are no peers connected

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ and this project adheres to
 ### Added
 
 - feat: implemented Net Api methods
-- feat: add automatic deployment of eoa from RPC if an account doesn't exist on chain
+- feat: add rpc docker image for hive testing
+- feat: add automatic deployment of eoa from RPC if an account doesn't exist on
+  chain
 - ci: decrease the rate which we send transactions in benchmark CI
 - fix: update the hive genesis utils for missing Kakarot requirements.
 - ci: use madara binary for benchmark CI
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,3 +27,4 @@ and this project adheres to
 - ci: cross-compile binaries to improve build time
 - test: update integration tests to use prepopulated Katana dumped state
 - ci: add `CHANGELOG.md` and enforce it is edited for each PR on `main`
+- feat: implemented Net Api methods
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- feat: implemented Net Api methods
 - feat: add automatic deployment of eoa from RPC if an account doesn't exist on chain
 - ci: decrease the rate which we send transactions in benchmark CI
 - fix: update the hive genesis utils for missing Kakarot requirements.
@@ -27,4 +28,3 @@ and this project adheres to
 - ci: cross-compile binaries to improve build time
 - test: update integration tests to use prepopulated Katana dumped state
 - ci: add `CHANGELOG.md` and enforce it is edited for each PR on `main`
-- feat: implemented Net Api methods
--- a/crates/eth-rpc/src/servers/net_rpc.rs
+++ b/crates/eth-rpc/src/servers/net_rpc.rs
@@ -3,7 +3,7 @@ use kakarot_rpc_core::client::constants::CHAIN_ID;
 use kakarot_rpc_core::client::errors::EthApiError;
 use reth_primitives::U64;
 use reth_rpc_types::PeerCount;
-
+use reth_network_api::PeersInfo;
 use crate::api::net_api::NetApiServer;
 
 /// The RPC module for the implementing Net api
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -27,3 +27,4 @@ and this project adheres to
 - ci: cross-compile binaries to improve build time
 - test: update integration tests to use prepopulated Katana dumped state
 - ci: add `CHANGELOG.md` and enforce it is edited for each PR on `main`
+- feat: implemented Net Api methods
```
bajpai244, line 30: Changelog should be bottom to up, i.e the most recent tweet should go up.

aniketpr01: Done!

crates/eth-rpc/src/servers/net_rpc.rs
```diff
@@ -23,12 +22,10 @@ impl NetApiServer for NetRpc {
     }
 
     fn peer_count(&self) -> Result<PeerCount> {
-        // TODO: replace jsonrpsee::types::ErrorObject with Provider::Error once NetRpc uses Provider
-        Err(EthApiError::<jsonrpsee::types::ErrorObject>::MethodNotSupported("eth_peerCount".to_string()).into())
+        Ok(PeerCount::Number(0))
```
bajpai244, line 25: Rather than having a hardcoded constant, better to have a lazy_static for this, like we have [here](https://github.com/kkrt-labs/kakarot-rpc/blob/50fc732165871d13b1acb2eb81778b75b14c0beb/crates/core/src/client/constants.rs#L93C5-L93C5).

So it can be:

```rust
lazy_static! {
  pub static ref PEER_COUNT: PeerCount  = PeerCount::Number(0) 
 }
```

bajpai244: and then you can use this lazy_static in place of the hardcoded value

aniketpr01: ahh yes, good suggestion, done!

crates/eth-rpc/src/servers/net_rpc.rs
```diff
@@ -23,12 +22,10 @@ impl NetApiServer for NetRpc {
     }
 
     fn peer_count(&self) -> Result<PeerCount> {
-        // TODO: replace jsonrpsee::types::ErrorObject with Provider::Error once NetRpc uses Provider
-        Err(EthApiError::<jsonrpsee::types::ErrorObject>::MethodNotSupported("eth_peerCount".to_string()).into())
+        Ok(PEER_COUNT.clone())
```
bajpai244, line 25: you can just do *PEER_COUNT.

aniketpr01: PeerCount does not implement Copy trait, so can't dereference it and has to use clone()

bajpai244:
@Eikix what do you think about this PR? @aniketpr01 is trying to implementation for two methods of[ the net api](https://www.quicknode.com/docs/ethereum/net_listening).

- peerCount: number of peers connected to the node, from my understanding in our case should zero
- listening: is it allowing for network connections to be made { from my understand p2p connection }, which I think should be false as well in our case as of now

Lemme know what you think.

crates/eth-rpc/src/servers/net_rpc.rs
```diff
@@ -23,12 +22,10 @@ impl NetApiServer for NetRpc {
...
     }
 
     fn listening(&self) -> Result<bool> {
-        // TODO: replace jsonrpsee::types::ErrorObject with Provider::Error once NetRpc uses Provider
-        Err(EthApiError::<jsonrpsee::types::ErrorObject>::MethodNotSupported("eth_listening".to_string()).into())
+        Ok(false)
```
Eikix, line 29: Can we add a comment why we're currently not listening?

aniketpr01: what should be documented for the comment? @bajpai244 @Eikix

Eikix:
```suggestion
        // Kakarot RPC currently does not support peer-to-peer connections
        Ok(false)
```

aniketpr01: Done!

crates/eth-rpc/src/servers/net_rpc.rs
bajpai244: @Eikix, I have no issues with what u are suggesting, it was just that I thought since we have constants for all these properties, then might be having one for this would also be consistent.

But, I don't see any issue with `PeerCount::Number(0)`, @aniketpr01 I think we can rollback to that, and if we need this property a second time, we can refactor back to a lazy_static!

bajpai244:
@aniketpr01 I guess you would also have to a rebase.

crates/eth-rpc/src/servers/net_rpc.rs
aniketpr01: Done!

Eikix (dismissed):
lgtm 🙏🙏💪💪💥

aniketpr01:
> @aniketpr01 you would also have to do a rebase.

Resolved!

Eikix (approved):
lgtm 💥

Issue #502, opened 2023-08-31
feat: update `nonce` to return managed nonce for contract accounts


ftupas:
# Feature Request

**Describe the Feature Request**

Currently, we return the protocol-level nonce both for EOA and Contract Accounts 

https://github.com/kkrt-labs/kakarot-rpc/blob/50fc732165871d13b1acb2eb81778b75b14c0beb/crates/core/src/client/mod.rs#L318-L333

**Describe Preferred Solution**

When this [pr](https://github.com/kkrt-labs/kakarot/pull/716) is merged, we need to do the following:
- update [nonce](https://github.com/kkrt-labs/kakarot-rpc/blob/50fc732165871d13b1acb2eb81778b75b14c0beb/crates/core/src/client/mod.rs#L318-L333) so that for EOA we take the native nonce and for Contract Accounts, the managed nonce
- add tests for the new nonce logic

PR #513, opened 2023-09-04
dev: update nonce for evm accounts

ftupas:
Time spent on this PR: 1

Resolves: #502 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [X] Build-related changes
- [ ] Testing

# What is the new behavior?

`nonce` method now returns the following:
- If the account is an EOA, return protocol-level nonce
- Otherwise if it's a contract account, we use managed nonce

- [ ] Yes
- [X] No


crates/core/tests/client.rs
```diff
@@ -317,4 +320,25 @@ mod tests {
...
+            client.nonce(plain_opcodes_eth_address, BlockId::from(BlockNumberOrTag::Latest)).await.unwrap();
+
+        let hash = execute_tx(&kakarot_test_env_ctx, "PlainOpcodes", "create", vec![U256::default(), count]).await;
+        client.transaction_receipt(hash).await.expect("create transaction failed");
+
+        let nonce_final =
+            client.nonce(plain_opcodes_eth_address, BlockId::from(BlockNumberOrTag::Latest)).await.unwrap();
+
+        // Then
+        assert_eq!(nonce_final, nonce_initial + count);
```
ftupas, line 342: This is currently failing

crates/core/src/client/config.rs
```diff
@@ -11,10 +11,21 @@ use url::Url;
...
+    let env_var = env_var(var_name)?;
+
+    FieldElement::from_hex_be(&env_var).map_err(|_| {
+        ConfigError::EnvironmentVariableSetWrong(format!(
+            "{} should be provided as a hex string, got {}",
+            var_name, env_var
+        ))
+    })
+}
+
```
ftupas, line 28: Added helper to reduce boiler plate

crates/core/src/test_utils/execution_helpers.rs
```diff
@@ -1,26 +1,23 @@
+use ethers::abi::Token;
 use ethers::signers::{LocalWallet, Signer};
-use reth_primitives::{Address, BlockId, H256, U256};
+use reth_primitives::{Address, BlockId, H256};
 
 use super::deploy_helpers::{create_eth_transfer_tx, create_raw_ethereum_tx, KakarotTestEnvironmentContext};
 use crate::client::api::KakarotEthApi;
 
-pub async fn execute_tx(env: &KakarotTestEnvironmentContext, contract: &str, selector: &str, args: Vec<U256>) -> H256 {
+pub async fn execute_tx(env: &KakarotTestEnvironmentContext, contract: &str, selector: &str, args: Vec<Token>) -> H256 {
```
ftupas, line 8: Use `Token` instead of `U256` so we can encode it

crates/core/tests/client.rs
```diff
@@ -25,7 +26,7 @@ mod tests {
 
     #[ctor]
     fn setup() {
-        let subscriber = FmtSubscriber::builder().with_max_level(tracing::Level::ERROR).finish();
+        let subscriber = FmtSubscriber::builder().with_max_level(tracing::Level::WARN).finish();
```
ftupas, line 29: Bump to `WARN` so we get some VM errors

crates/core/tests/client.rs
ftupas: This test is now passing

crates/core/tests/client.rs
```diff
@@ -278,6 +292,9 @@ mod tests {
         let balance =
             client.balance(ethereum_address_to_deploy, BlockId::Number(BlockNumberOrTag::Latest)).await.unwrap();
         assert_eq!(balance, U256::ZERO);
+
+        let nonce = client.nonce(ethereum_address_to_deploy, BlockId::Number(BlockNumberOrTag::Latest)).await.unwrap();
+        assert_eq!(nonce, U256::ZERO);
```
ftupas, line 297: not sure if I should make a separate `test_get_nonce_eoa` so I just added  the `nonce` check here

crates/core/src/mock/mock_starknet.rs
```diff
@@ -52,6 +58,7 @@ impl Serialize for AvailableFixtures {
         match self {
             AvailableFixtures::ComputeStarknetAddress => serializer.serialize_str("kakarot_computeStarknetAddress"),
             AvailableFixtures::GetEvmAddress => serializer.serialize_str("kakarot_getEvmAddress"),
+            AvailableFixtures::GetImplementation => serializer.serialize_str("kakarot_getImplementation"),
```
ClementWalter, line 61: not sure to follow, kakarot has no `get_implementation` entrypoint

crates/core/src/client/config.rs
```diff
@@ -61,19 +72,35 @@ pub struct StarknetConfig {
     pub kakarot_address: FieldElement,
     /// Proxy account class hash.
     pub proxy_account_class_hash: FieldElement,
+    /// EOA class hash.
+    pub externally_owned_account_class_hash: FieldElement,
+    /// Contract Account class hash.
+    pub contract_account_class_hash: FieldElement,
```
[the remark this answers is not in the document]
ftupas: Definitely agree, better to remove them from the config and directly get it from the contract 👍

crates/core/src/mock/mock_starknet.rs
ftupas: gotcha, will change this should to account instead of kakarot

crates/core/src/client/config.rs
```diff
@@ -11,10 +11,21 @@ use url::Url;
...
+fn env_var(name: &str) -> Result<String, ConfigError> {
     std::env::var(name).map_err(|_| ConfigError::EnvironmentVariableMissing(name.into()))
 }
 
+fn field_element_from_env(var_name: &str) -> Result<FieldElement, ConfigError> {
+    let env_var = env_var(var_name)?;
+
+    FieldElement::from_hex_be(&env_var).map_err(|_| {
+        ConfigError::EnvironmentVariableSetWrong(format!(
+            "{} should be provided as a hex string, got {}",
```
Eikix, line 23: try running `cargo clippy --  -W clippy::pedantic -W clippy::nursery` I think it has a slightly more concise way to write format! inside an error

Eikix, line 23: Isn't there an error case where the input is a valid hex string, but bigger than 31 bytes or 32 bytes i.e. longer than 65 or 66 chars?

crates/core/src/client/mod.rs
```diff
@@ -322,26 +333,41 @@ impl<P: Provider + Send + Sync + 'static> KakarotEthApi<P> for KakarotClient<P>
     }
 
     /// Returns the nonce for a given ethereum address
+    /// if it's an EOA, use native nonce and if it's a contract account, use managed nonce
```
Eikix, line 336: nice

crates/core/src/client/mod.rs
```diff
@@ -322,26 +333,41 @@ impl<P: Provider + Send + Sync + 'static> KakarotEthApi<P> for KakarotClient<P>
...
-                    code: MaybeUnknownErrorCode::Known(StarknetError::ContractNotFound),
-                    ..
-                }) => Ok(U256::from(0)),
-                _ => Err(EthApiError::from(err)),
-            })
+        // Get the implementation of the account
+        let account = KakarotAccount::new(starknet_address, &self.starknet_provider);
+        let implementation = match account.get_implementation(&starknet_block_id).await {
+            Ok(class_hash) => class_hash,
+            Err(_) => return Ok(U256::from(0)), // Return 0 if the account doesn't exist
```
Eikix, line 347: can use `unwrap_or` here I think

crates/core/src/contracts/contract_account.rs
```diff
@@ -58,4 +58,23 @@ impl<'a, P: Provider + Send + Sync> ContractAccount<'a, P> {
         let value = Into::<U256>::into(low) + (Into::<U256>::into(high) << 128);
         Ok(value)
     }
+
+    /// Returns the nonce of the contract account.
```
Eikix, line 62: I think since this method is a bit special, it deserves more context in the comment, 
- something explaining that there are two types on accounts in EVM, EOA and Contract Account. That EOA nonce is handled by Starknet protocol and that Contract Account nonce is handled by a dedicated storage 
- this function fetches the nonce in storage

crates/core/src/contracts/contract_account.rs
```diff
@@ -58,4 +58,23 @@ impl<'a, P: Provider + Send + Sync> ContractAccount<'a, P> {
...
+        if result.len() != 1 {
+            return Err(DataDecodingError::InvalidReturnArrayLength {
+                entrypoint: "get_nonce".into(),
+                expected: 1,
+                actual: result.len(),
+            }
+            .into());
+        }
+
+        Ok(U256::from_be_bytes(result[0].to_bytes_be()))
```
Eikix, line 78: Cleaner to go through the felt252wrapper imo here, since it's our go-to wrapper for type conversion

Eikix (changes requested):
first review before getting into the plane, nice PR:)!

crates/core/src/contracts/contract_account.rs
ftupas: Gotcha adding this in the docstring

ftupas (commented):
Thanks for the review @Eikix

crates/core/src/contracts/contract_account.rs
```diff
@@ -79,6 +79,6 @@ impl<'a, P: Provider + Send + Sync> ContractAccount<'a, P> {
             .into());
         }
 
-        Ok(U256::from_be_bytes(result[0].to_bytes_be()))
+        Ok(Into::<Felt252Wrapper>::into(result[0]).into())
```
Eikix, line 82: `FeltWrapper252::from(result[0]).into()`

crates/core/src/contracts/account.rs
```diff
@@ -54,6 +54,30 @@ pub trait Account<'a, P: Provider + Send + Sync + 'a> {
         // TODO: Remove Manual Decoding
         Ok(vec_felt_to_bytes(bytecode[1..].to_vec()))
     }
+
+    /// Returns the class hash of account implementation of the contract.
+    async fn get_implementation(&self, block_id: &BlockId) -> Result<FieldElement, EthApiError<P::Error>> {
```
Eikix, line 59: should we call it `implementation()` to respect our no 'get' policy?

crates/core/src/contracts/contract_account.rs
```diff
@@ -58,4 +58,27 @@ impl<'a, P: Provider + Send + Sync> ContractAccount<'a, P> {
         let value = Into::<U256>::into(low) + (Into::<U256>::into(high) << 128);
         Ok(value)
     }
+
+    /// Returns the nonce of the contract account.
+    /// In Kakarot EVM, there are two types of accounts: EOA and Contract Account.
+    /// EOA nonce is handled by Starknet protocol.
+    /// Contract Account nonce is handled by Kakarot through a dedicated storage, this function
+    /// returns that storage value.
+    pub async fn nonce(&self, block_id: &BlockId) -> Result<U256, EthApiError<P::Error>> {
```
Eikix, line 67: is this tested?

crates/core/src/contracts/account.rs
Eikix, line 59: is this thoroughly tested?

crates/core/src/mock/mock_starknet.rs
```diff
@@ -38,7 +42,9 @@ impl From<AvailableFixtures> for JsonRpcMethod {
     fn from(value: AvailableFixtures) -> Self {
         match value {
             AvailableFixtures::Other(method) => method,
-            AvailableFixtures::ComputeStarknetAddress | AvailableFixtures::GetEvmAddress => JsonRpcMethod::Call,
+            AvailableFixtures::ComputeStarknetAddress
```
Eikix, line 45: Isn't `get_nonce` also a call?
Maybe the contract_account.nonce() method deserves to be unitarily tested

Eikix (commented):
Should there be an overall integration test for this feature? Using the testsequencer flow

crates/core/src/mock/mock_starknet.rs
ftupas: Gotcha will add a unit test

crates/core/src/contracts/account.rs
ftupas: Will update 🙂

ftupas:
> Should there be an overall integration test for this feature? Using the testsequencer flow

Yeah [here](https://github.com/ftupas/kakarot-rpc/blob/0f4a260c8e8e77f637b0930052015f34c87ce533/crates/core/tests/client.rs#L346) we are testing for contract account and added the nonce test [here](https://github.com/ftupas/kakarot-rpc/blob/0f4a260c8e8e77f637b0930052015f34c87ce533/crates/core/tests/client.rs#L275) for eoa 🤙

crates/core/src/client/config.rs
```diff
@@ -61,19 +67,35 @@ pub struct StarknetConfig {
...
+        externally_owned_account_class_hash: FieldElement,
+        contract_account_class_hash: FieldElement,
+    ) -> Self {
+        StarknetConfig {
+            network,
+            kakarot_address,
+            proxy_account_class_hash,
+            externally_owned_account_class_hash,
+            contract_account_class_hash,
+        }
```
ClementWalter, line 90: though I do see in some rust example this `new` function that does nothing more than creating the struct with all the param, I still don't understand why we do so. any hint?

Eikix: iirc it is if struct fields are private then you need to use new or default

ftupas: adding to @Eikix, it can also be for encapsulation, from my understanding. Even with public fields, if you have a `new` function to create instances, you can abstract away the defails of instantiation i.e. can change the internal details without changing the API

crates/core/src/contracts/tests/account.rs
```diff
@@ -1,15 +1,53 @@
...
+        let fixtures = fixtures(vec![AvailableFixtures::GetImplementation]);
+        let starknet_provider = Arc::new(mock_starknet_provider(Some(fixtures)));
+        let contract_account = ContractAccount::new(*ABDEL_STARKNET_ADDRESS, &starknet_provider);
+
+        // When
+        let implementation = contract_account.implementation(&BlockId::Tag(BlockTag::Latest)).await.unwrap();
+
+        // Then
+        assert_eq!(
+            FieldElement::from_hex_be("0x050de736bbc22be1cdb0c741afc80d01b6f5d1c248034a4cbe6f0eba6bde3051").unwrap(),
```
Eikix, line 47: how come this isnt "0x04ea3da1315fd04212a6f08f2ab7c9ef40cfb1cefa86633d9f9f308a1275367d";

ftupas: Fixed, the eoa and contract account class hashes got swapped

crates/core/src/client/mod.rs
```diff
@@ -322,26 +333,41 @@ impl<P: Provider + Send + Sync + 'static> KakarotEthApi<P> for KakarotClient<P>
...
-                    code: MaybeUnknownErrorCode::Known(StarknetError::ContractNotFound),
-                    ..
-                }) => Ok(U256::from(0)),
-                _ => Err(EthApiError::from(err)),
-            })
+        // Get the implementation of the account
+        let account = KakarotAccount::new(starknet_address, &self.starknet_provider);
+        let implementation = match account.implementation(&starknet_block_id).await {
+            Ok(class_hash) => class_hash,
+            Err(_) => return Ok(U256::from(0)), // Return 0 if the account doesn't exist
```
Eikix, line 347: are there other errors that need to bubble up? 
like if account doesn't exist then 0 but what if we fail our call for another reason

Eikix (commented):
question

crates/core/src/client/mod.rs
ftupas: Right we need to handle `ContractNotFound` and return 0, otherwise we propagate the error, will fix

Eikix (approved):
lgtm 🫡🫡🙏

Eikix (approved):
lgtm

Issue #547, opened 2023-09-11
dev: refactor all test utils to a seperate crate `test_utils`


bajpai244:
In the current implementation, we seem to have [hive-utils](https://github.com/kkrt-labs/kakarot-rpc/tree/main/crates/hive-utils) as a seperate crate, while [test_utils](https://github.com/kkrt-labs/kakarot-rpc/tree/main/crates/core/src/test_utils) live inside the `core` crate.

It makes more sense, to refactor a common `test_utils` crate, and put both of this inside.

**Describe Preferred Solution**

- refactor [test_utils](https://github.com/kkrt-labs/kakarot-rpc/tree/main/crates/core/src/test_utils) & [hive-utils](https://github.com/kkrt-labs/kakarot-rpc/tree/main/crates/hive-utils) to a common crate `test_utils`.

PR #558, opened 2023-09-12
dev: refactor hive-utils crate and kakarot_rpc::test_utils to test_utils crate

bajpai244:
Resolves: #547 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

The [hive-utils crate](https://github.com/kkrt-labs/kakarot-rpc/tree/main/crates/hive-utils) and [kakarot_rpc::test_utils](https://github.com/kkrt-labs/kakarot-rpc/blob/main/crates/eth-rpc/src/test_utils.rs) have been refactored to a seperate test_utils crate.

- [ ] Yes
- [x] No


crates/test-utils/Cargo.toml
```diff
@@ -13,11 +13,17 @@ license.workspace = true
 
 [dependencies]
 kakarot-rpc-core={ path = "../core" }
+kakarot-rpc= { path = "../eth-rpc" }
```
ClementWalter, line 16: I thought that the test utils weren't supposed to depend on the RPC any more

bajpai244: @ClementWalter I can remove this dependency, but that means moving [kakarot_rpc::test_utils](https://github.com/kkrt-labs/kakarot-rpc/blob/main/crates/eth-rpc/src/test_utils.rs).

So, if you think that will work then lemme know!

bajpai244: @ClementWalter as discussed in the chat, I don't think we should move it back, we need a better refactor of the code to avoid this!

We can move ahead with it, but then later on focus to refactor the codebase so that such coupling doesn't exist.

ClementWalter (approved):

Issue #641, opened 2023-11-07
feat: add health check path


apoorvsadana:
# Feature Request

**Describe the Feature Request**

Having a health check path is generally helpful when deploying with Cloud providers. The current RPC only supports `POST` and `OPTIONS` requests and AWS only allows GET requests for Health checks in target groups. I am assuming this might also be the case with other providers.

**Describe Preferred Solution**

Add a GET `/health` endpoint which

- Checks if the RPC_URL is still online and returns a boolean. Sample response
```
{ "rpc_online": true }
```

TAdev0:
Hi @Eikix, m

TAdev0:
Hello @apoorvsadana  @Eikix ,
I spent a few times trying to figure out what should be done concretely. This is my first time dealing with infra.
Could you provide me a slightly more detailed explanation about how to handle this health check? Should i add a function in `eth_rpc.rs` ?
Thank you.

Eikix:
> Hello @apoorvsadana @Eikix , I tried a few times to figure out what should be done concretely. This is my first time dealing with infra. Could you provide me a slightly more detailed explanation about how to handle this health check? Should i add a function in `eth_rpc.rs` ? Thank you.

Hey @TAdev0 thanks for taking the time.
I haven't given this a lot of thought, but we'd probably need to add a file in the eth-rpc crate, where all the different servers are declared: eth-api, alchemy-api etc., we could add our own there, with its Health method.

Let's see what @apoorvsadana suggests, but the method should live outside of the ethereum rpc spec, but as an "exposed" method -> the only way is to add a new server

TAdev0:
implementing
```
  async fn health(&self) -> Result<bool> {
        Ok(true)
    }
 ```
 would return true only if the RPC is true

TAdev0:
@Eikix  ok i'll try to do that.
What is the core difference between API and Servers folders? This feature doesn't modification in the server folder files?

Eikix:
> @Eikix ok i'll try to do that. Thank you for clarification !

I'm def no expert either at infra / network,
You may want to check how other clients  and full node implement their health endpoint

TAdev0:
it seems that :
- reth Net API only exposes `version`, `peer_count` and `is_listening` and doesn't have a health check. 
- reth Eth API doesn't have a health check either

@apoorvsadana  did you think of any special way to implement it?

Eikix:
> it seems that :
> 
[...]

PR #652, opened 2023-11-16
FEAT: add health check path

TAdev0:
Time spent on this PR: 30min

Resolves: #<Issue number>
#641 

# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

This PR exposes a health check through Net API and and eth API to check that the RPC URL is currently online.

I also added missing  comments for `block_number` function in net_api.rs file.

- [ ] Yes
- [X] No


crates/eth-rpc/src/api/eth_api.rs
```diff
@@ -9,6 +9,11 @@ use reth_rpc_types::{
 #[rpc(server, namespace = "eth")]
 #[async_trait]
 pub trait EthApi {
+    /// Returns true if RPC_URL is online
+    #[method(name = "health")]
```
Eikix, line 13: I don't think this method should be in the EthApi, 
The EthApi is a documented and quite mature specification:
https://ethereum.github.io/execution-apis/api-documentation/

We should implement it in a custom Api.
Suggestion:
- Delete this function in EthApi (eth_api.rs)
- Delete the implementation of the trait in server/eth_rpc.rs


This will still make net_health reachable, but eth_health, not

crates/eth-rpc/src/servers/net_rpc.rs
```diff
@@ -30,4 +30,8 @@ impl NetApiServer for NetRpc {
         // Kakarot RPC currently does not support peer-to-peer connections
         Ok(false)
     }
+
+    fn health(&self) -> Result<bool> {
```
Eikix, line 34: Apoorv, is this satisfactory for a health endpoint?

Eikix:
Hey, after talking with Apoorv, it'd be better to check that the underlying Starknet RPC server is also working,
A good way to do this is to call the underlying Starknet provider's starknet_blockNumber, and check if it resolves.
If it's an error, return an error,
if it's defined (i.e. returns a number), then you can return Ok(true)

One way to do this is to have `impl NetApiServer for NetRpc` and `NetRpc` wrap a KakarotClient, just like EthApi.
i.e. go from 
pub struct NetRpc {}

impl NetRpc {
    pub fn new() -> Self {
        Self {}
    }
}

to

pub struct NetRpc<P: Provider + Send + Sync> {
    pub kakarot_client: Arc<KakarotClient<P>>,
}

impl<P: Provider + Send + Sync> NetRpc<P> {
    pub fn new(kakarot_client: Arc<KakarotClient<P>>) -> Self {
        Self { kakarot_client }
    }
}

And then you can properly implement your desired Health new function

crates/eth-rpc/src/servers/net_rpc.rs
```diff
@@ -30,4 +33,12 @@ impl NetApiServer for NetRpc {
         // Kakarot RPC currently does not support peer-to-peer connections
         Ok(false)
     }
+
+    fn health(&self) -> Result<bool, io::Error> {
+        // Calls `starknet_blockNumber` method to check if it resolves
+        match self.kakarot_client.starknet_blockNumber() {
```
Eikix, line 39: I think this method does not exist, maybe you can check this by adding a health call at the beginning on integration.rs

I think the correct impl is:
`self.kakarot_client.starknet_provider().block_number()`

Eikix: Moreover, you'll need to handle the possible error, maybe with map_err for now, 
in the following way:
`self.kakarot_client.starknet_provider().block_number().await.map_err(EthApiError::from)?;`

crates/eth-rpc/src/servers/net_rpc.rs
```diff
@@ -30,4 +33,12 @@ impl NetApiServer for NetRpc {
         // Kakarot RPC currently does not support peer-to-peer connections
         Ok(false)
     }
+
+    fn health(&self) -> Result<bool, io::Error> {
```
Eikix, line 37: I don't think the health error should be an io::Error,
I think it can remain a Result<bool>
Meaning that there is an implicit Error that is of type jsonrpsee::Error

Leave it as
`fn health(&self) -> Result<bool>`

crates/eth-rpc/src/servers/net_rpc.rs
TAdev0: @Eikix  i just updated the PR. Again, not sure about my code.
Is it normal that i need to add `--no-verify` when i commit with git, otherwise i have issues committing. Indeed, with my rust extension, i can see some lines of code in core crate are not ok.

TAdev0: I checked integration.rs file but i'm not sure about what i should do in there exactly. Could you give me some, more tips ?

Eikix: > @Eikix i just updated the PR. Again, not sure about my code. Is it normal that i need to add `--no-verify` when i commit with git, otherwise i have issues committing. Indeed, with my rust extension, i can see some lines of code in core crate are not ok. (more precisely, mod.rs and transaction.rs)

I think it is the precommit hook that returns an error, saying some of your code is not respecting the linter / compilation.

Eikix: > I checked integration.rs file but i'm not sure about what i should do in there exactly. Could you give me some, more tips ?

I think you'll need to slightly modify KakarotRpcModuleBuilder::new to have NetApi take a reference to KakarotClient, akin to other RPC modules

Then at the beginning of integration.rs, right before `let wallet= ...`
You can add a reqwest post request to net_health endpoint, 
endpoint address: `format!("http://localhost:{}", server_addr.port())`
endpoint selector: `net_health`
http word: `post`

take reference to https://docs.rs/reqwest/latest/reqwest/#json

Eikix: you should:
- rebase on main 
- run `make setup`, note that you may need to copy .env.example into your .env 
- then run test

TAdev0: everything works perfectly now, thanks!

TAdev0: @Eikix  i was trying to push my new commit but i'm having an issue with the pre hook i activated again : 

```
tristan@Tristans-MBP kakarot-rpc % git commit -m healthCheck
+cargo clippy --all -- -D warnings
    Checking kakarot-rpc-core v0.1.0 (/Users/tristan/Desktop/Dev/kakarot-rpc/crates/core)
error[E0308]: mismatched types
   --> crates/core/src/client/mod.rs:226:48
    |
226 |         let storage = contract_account.storage(&Uint256 { low: key_low.into(), high: key_high.into() }).call().await?;
    |                                        ------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&FieldElement`, found `&Uint256`
    |                                        |
    |                                        arguments to this method are incorrect
    |
    = note: expected reference `&starknet_crypto::FieldElement`
               found reference `&client::Uint256`
note: method defined here
   --> crates/core/src/client/mod.rs:51:1
    |
51  | abigen_legacy!(ContractAccount, "./artifacts/contract_account.json");
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the macro `abigen_legacy` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0308`.
error: could not compile `kakarot-rpc-core` (lib) due to previous error

```

This is not related to my modifications. This issue is in the core crate, mod.rs file. What am i supposed to do?

Eikix: Weird! Did you rebase on main?

TAdev0: yes i did

TAdev0: The codebase of kakarot RPC is the same and there is this mismatch of data type. Does this mean the codebase of main has an inconstancy?  how could this be possible with the pre commit hook ?

Eikix: Yes I agree it's weird! It shouldn't be the case, let me check

Eikix: The code in Kakarot RPC is up to date, did you try to pull main and then rebase ?

TAdev0: i think i'm good now ! thank you

Eikix:
I suspect your rebase on main failed since you're getting all the diffs :(

You should ->
Pull kkrt-labs/main
checkout your healthCheck branch
git rebase kkrt-labs/main

(this is pseudo code but you get the jist of it)

TAdev0:
@Eikix  i'm so confused about git... i dont understand i did git rebase main. My repo is the same as the main but with my extra code added, thats all. Will try to do what you just said. thanks.. sorry for that

TAdev0:
@Eikix  may i create a new branch to make it easier? i can't find my way tbh

PR #724, opened 2024-01-23
Use new env var for defining kakarot's rpc url

d-roak:
@Eikix knows
- Use new env var for defining kakarot's rpc url
- Run katana as daemon
- Force block sealing in katana

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-rpc/724)


Eikix (approved):
lgtm

PR #752, opened 2024-02-02
add health endpoint

apoorvsadana:
1. Add a get /health endpoint
2. Add poetry version to Dockerfile as it installs an older version without that

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-rpc/752)


Eikix (approved):
lgtm

PR #929, opened 2024-04-05
Reorder operations in `EthDataProvider` `block` method

tcoratger:
Time spent on this PR: 20 min

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

Currently, in the `block` function of `EthDataProvider`, the block transactions were retrieved before checking if the `withdrawals_root` was correct. This operation, which is quite expensive, is unnecessary to carry out in the case where we know in advance that the `withdrawals_root` is false and that an error will therefore have to be returned.

The transaction query operation for the block is therefore deported to the end of the function, once we are sure that the `withdrawals_root` is valid or `None`.

- [ ] Yes
- [X] No


greged93 (approved):
lgtm

PR #935, opened 2024-04-08
Implement `CollectionName` enum to store db collection names

tcoratger:
Time spent on this PR: 20 min

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [X] Code style update (formatting, renaming)
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

To avoid any typos or errors when writing the names of the collections in the Mongo DB database, a `CollectionName` enum is created and it contains an `as_str` method in its implementation which allows you to find the name of all collections in a Rust idiomatic way.

- [ ] Yes
- [X] No


greged93 (commented):
I like this idea, but what do you think of the following:
- Add a trait which also to convert a type to string (as a static function, not a method).
```rust
   pub trait ToStringType {
      fn to_string() -> String;
   }
```
- Implement this on the `StoredXXX` types.
- Update the Mongo wrapper to take
```rust
pub async fn get_one<T>(
        &self,
        filter: impl Into<Option<Document>>,
        sort: impl Into<Option<Document>>,
    ) -> DatabaseResult<Option<T>>
    where
        T: DeserializeOwned + ToStringType + Unpin + Send + Sync,
```
Then we can just do in the body of the function
```rust
let find_one_option = FindOneOptions::builder().sort(sort).build();
let collection = self.0.collection::<T>(T::to_string());
let result = collection.find_one(filter, find_one_option).await?;
Ok(result)
```

greged93 (approved):

Issue #970, opened 2024-04-15
dev: use reth `TryFrom` impl to convert blocks and headers


tcoratger:
**Describe Preferred Solution**

After next reth release and integration of https://github.com/paradigmxyz/reth/pull/7660#event-12476448733

- `rpc_to_primitive_header` method 

https://github.com/kkrt-labs/kakarot-rpc/blob/d2128cf23e8981284e577cb00dcbbf9464f3beb1/src/models/block.rs#L71-L104

can be removed and replaced by reth native `TryFrom` implementation call.

- `rpc_to_primitive_block` method

https://github.com/kkrt-labs/kakarot-rpc/blob/d2128cf23e8981284e577cb00dcbbf9464f3beb1/src/models/block.rs#L106-L136

can be removed and replaced by reth native `TryFrom` implementation call.

bhavyagosai:
Hey! I think I can take this up if its up for grabs?

tcoratger:
> Hey! I think I can take this up if its up for grabs?

Sure just assigned

tcoratger:
For your information https://github.com/kkrt-labs/kakarot-rpc/pull/1004 merge should unlock the situation here.

bhavyagosai:
yes yes I saw the release has happened, working on it right now 😅

PR #1007, opened 2024-04-23
refactor: Use `reth` `try_from` impl to convert blocks and headers

bhavyagosai:
Time spent on this PR: 1hr (I'm relatively new to rust)

Resolves: #970 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

- [ ] Yes
- [x] No

### DIFF
--- a/src/eth_provider/provider.rs
+++ b/src/eth_provider/provider.rs
@@ -1,5 +1,4 @@
 use crate::eth_provider::starknet::kakarot_core::account_contract::BytecodeOutput;
-use crate::models::block::rpc_to_primitive_header;
 use alloy_rlp::{Decodable, Encodable};
 use async_trait::async_trait;
 use auto_impl::auto_impl;
@@ -11,7 +10,8 @@ use mongodb::bson::doc;
 use reth_primitives::constants::EMPTY_ROOT_HASH;
 use reth_primitives::serde_helper::{JsonStorageKey, U64HexOrNumber};
 use reth_primitives::{
-    Address, BlockId, BlockNumberOrTag, Bytes, TransactionSigned, TransactionSignedEcRecovered, B256, U256, U64,
+    Address, BlockId, BlockNumberOrTag, Bytes, Header as PrimitiveHeader, TransactionSigned,
+    TransactionSignedEcRecovered, B256, U256, U64,
 };
 use reth_rpc_types::{
     Block, BlockHashOrNumber, BlockTransactions, FeeHistory, Filter, FilterChanges, Header, Index, RichBlock,
@@ -826,7 +826,8 @@ where
 
         // This is how reth computes the block size.
         // `https://github.com/paradigmxyz/reth/blob/v0.2.0-beta.5/crates/rpc/rpc-types-compat/src/block.rs#L66`
-        let size = rpc_to_primitive_header(header.clone())?.length();
+        let size =
+            PrimitiveHeader::try_from(header.clone()).ok().ok_or(EthereumDataFormatError::PrimitiveError)?.length();
         Ok(Some(
             Block {
                 header,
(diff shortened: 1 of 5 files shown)


bhavyagosai:
not able to figure out why `test_rpc_to_primitive_block` is failing.

What I essentially changed was remove the `rpc_to_primitive_block(block).unwrap()` method used there with `Block::try_from(block).unwrap();`

src/eth_provider/provider.rs
```diff
@@ -826,7 +826,8 @@ where
 
         // This is how reth computes the block size.
         // `https://github.com/paradigmxyz/reth/blob/v0.2.0-beta.5/crates/rpc/rpc-types-compat/src/block.rs#L66`
-        let size = rpc_to_primitive_header(header.clone())?.length();
+        let size =
+            PrimitiveHeader::try_from(header.clone()).ok().ok_or(EthereumDataFormatError::PrimitiveError)?.length();
```
tcoratger, line 830: Here:
- As `PrimitiveHeader` is only used here please use `reth_primitives::Header` instead and remove it from the import list.
- I think we should use an approach like `PrimitiveHeader::try_from(header.clone()).map_err(|_| EthereumDataFormatError::PrimitiveError)?.length()` (or something like that) because we cannot suppose that the result of the conversion will be fine and we must handle error properly.

src/eth_rpc/servers/debug_rpc.rs
```diff
@@ -47,7 +45,10 @@ impl<P: EthereumProvider + Send + Sync + 'static> DebugApiServer for DebugRpc<P>
         };
         let mut raw_block = Vec::new();
         if let Some(block) = block {
-            let block = rpc_to_primitive_block(block.inner).map_err(EthApiError::from)?;
+            let block = Block::try_from(block.inner)
+                .ok()
+                .ok_or(EthereumDataFormatError::PrimitiveError)
```
tcoratger, line 50: why using `ok` and `ok_or` can't we just map the error

src/eth_rpc/servers/debug_rpc.rs
```diff
@@ -47,7 +45,9 @@ impl<P: EthereumProvider + Send + Sync + 'static> DebugApiServer for DebugRpc<P>
         };
         let mut raw_block = Vec::new();
         if let Some(block) = block {
-            let block = rpc_to_primitive_block(block.inner).map_err(EthApiError::from)?;
+            let block = Block::try_from(block.inner)
+                .map_err(|_| EthereumDataFormatError::PrimitiveError)
+                .map_err(EthApiError::from)?;
```
tcoratger, line 50: I think we can do both steps into a single `map_err` no?

bhavyagosai: merged!

tcoratger:
I think here https://github.com/kkrt-labs/kakarot-rpc/blob/b0929dd54221dd3a924f246b9ce00871e354a8a3/src/models/block.rs#L199 you need to provide a random or empty access list but not `None` because conversion to primitive type doesn't support `None` values, see https://github.com/paradigmxyz/reth/blob/4f81f3acc9fdc56b18444c6754f852b7060c57ee/crates/primitives/src/transaction/mod.rs#L682

bhavyagosai:
@tcoratger I will pass a default access list from `reth_rpc_types::AccessList::default()`. That is okay right?

tcoratger:
> @tcoratger I will pass a default access list from `reth_rpc_types::AccessList::default()`. That is okay right?

Yes I think it works

tcoratger:
As withdrawals are not supported, this 

https://github.com/kkrt-labs/kakarot-rpc/blob/b0929dd54221dd3a924f246b9ce00871e354a8a3/src/models/block.rs#L215

should be a `None`

tcoratger:
Pending failing test and then we can merge, you can test locally to debug properly, you should have a withdrawals root that is not `None` in the test.

tcoratger (approved):

bhavyagosai:
alright! I was trying to figure out stuff on local and your commits after you pushed them for fixing the test. Reviewing this so I can understand the test properly haha. Thank you for the opportunity!

PR #1170, opened 2024-06-08
dev: move TRANSACTION_MAX_RETRIES to env

jbmkahdeksan:
Time spent on this PR: 2.5 hours

Resolves: #1162 

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

`TRANSACTION_MAX_RETRIES` is now an env variable.
env.example, and various Docker Compose and Dockerfiles were updated.

- [ ] Yes
- [x] No


jbmkahdeksan:
One doubt I had was regarding the new location of TRANSACTION_MAX_RETRIES. Initially it was a constant, now it is an environment variable. So possible places I thought correct were:
- `src/eth_provider/starknet/kakarot_core.rs` because many env variables are declared and used here.
- `/src/eth_provider/provider.rs` because `TRANSACTION_MAX_RETRIES` is used here.

Opted to leave it where it was, but happy to receive feedback about this.

src/eth_provider/constant.rs
```diff
@@ -20,7 +20,10 @@ pub const ADDRESS_HEX_STRING_LEN: usize = 40;
 /// Starknet Modulus: 0x800000000000011000000000000000000000000000000000000000000000001
 pub const STARKNET_MODULUS: U256 = U256::from_limbs([0x1, 0, 0, 0x0800_0000_0000_0011]);
 /// Maximum number of times a transaction can be retried
-pub const TRANSACTION_MAX_RETRIES: u8 = 10;
+pub const TRANSACTION_MAX_RETRIES: u8 = u8::from_str(
+    &std::env::var("TRANSACTION_MAX_RETRIES")
+        .unwrap_or_else(|_| panic!("Missing environment variable TRANSACTION_MAX_RETRIES"))
+).expect("failing to parse TRANSACTION_MAX_RETRIES");
```
tcoratger, line 26: You need to import 

```rust
use std::str::FromStr;
```

See compilation error in the CI

jbmkahdeksan: oh!
didn't noticed, will fix it!
btw, could you address the other comment. I am afraid I discovered another error saying: "cannot call non-const fun <from_str> in constants"
so I will have to move TRANSACTION_MAX_ENTRIES to another file

tcoratger: Yes you can use lazy_static for that, see the example of `RETRY_TX_INTERVAL` because here at the moment you are declaring `TRANSACTION_MAX_RETRIES` as a pub const but you are using `from_str` which is not a const function and this is forbidden by Rust to do this.

tcoratger:
@jbmkahdeksan Please fix lints :)

tcoratger (approved):

Issue #1187, opened 2024-06-13
dev: cleanup indexer `transform` function


tcoratger:
**Describe the enhancement request**

At the moment, in the indexer, the `transform` function is responsible for transforming the transaction data in order to index them properly. See https://www.apibara.com/docs/indexers#data-transformation for further information on this.

This function, with time and the addition of features, has become quite complex and not very Typescript idiomatic. As per https://github.com/kkrt-labs/kakarot-rpc/pull/1145/files#r1638720404, we should do a general cleanup of this function and make it more Typescript idiomatic in general.

tcoratger:
> Hi, can I take this?

Assigned

PR #1226, opened 2024-06-22
cleanup indexer transform function

cwastche:
Time spent on this PR: 2h30

Resolves: #1187 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

- Changed return statements to array filters
- Moved object mutations from map to forEach

- [ ] Yes
- [x] No

### DIFF
--- a/indexer/src/main.ts
+++ b/indexer/src/main.ts
@@ -95,10 +95,10 @@ export default async function transform({
     // This is typically the case if there are multiple Kakarot contracts on the same chain.
     .filter((event) => isKakarotTransaction(event.transaction))
     // Skip if the transaction_executed event contains "eth validation failed".
-    .filter((event) => !ethValidationFailed(event))
+    .filter((event) => !ethValidationFailed(event.event))
     .map((event) => ({
       event: event,
-      typedEthTx: toTypedEthTx(event.transaction),
+      typedEthTx: toTypedEthTx({ transaction: event.transaction }),
     }))
     // Can be null if:
     // 1. The transaction is missing calldata.
@@ -107,10 +107,10 @@ export default async function transform({
     // 4. The chain id is not encoded in the v param of the signature for a
     //    Legacy transaction.
     // 5. The deserialization of the transaction fails.
-    .filter((eventExtended) => eventExtended.typedEthTx)
+    .filter((eventExtended) => eventExtended.typedEthTx !== null)
     .map((eventExtended) => {
       const ethTx = typedTransactionToEthTx({
-        typedTransaction: eventExtended.typedEthTx,
+        typedTransaction: eventExtended.typedEthTx!,
         receipt: eventExtended.event.receipt,
         blockNumber,
         blockHash,
@@ -121,7 +121,11 @@ export default async function transform({
     })
     // Can be null if:
     // 1. The typed transaction if missing a signature param (v, r, s).
-    .filter((eventExtended) => eventExtended.ethTx)
+    .filter((
+      eventExtended,
+    ): eventExtended is typeof eventExtended & {
+      ethTx: NonNullable<typeof eventExtended.ethTx>;
+    } => eventExtended.ethTx !== null)
     .map((eventExtended) => {
       const ethLogs = eventExtended.event.receipt.events
         .map((e) =>
@@ -152,7 +156,7 @@ export default async function transform({
       const ethReceipt = toEthReceipt({
         transaction: eventExtended.ethTx,
         logs: ethLogsIndexed,
-        event: eventExtended.event,
+        event: eventExtended.event.event,
         cumulativeGasUsed,
         blockNumber,
         blockHash,
@@ -197,7 +201,7 @@ export default async function transform({
     .map((eventExtended) =>
       createTrieData({
         transactionIndex: Number(eventExtended.ethTx.transactionIndex),
-        typedTransaction: eventExtended.typedEthTx,
+        typedTransaction: eventExtended.typedEthTx!,
         receipt: eventExtended.ethReceipt,
       })
     )
@@ -248,13 +252,13 @@ export default async function transform({
       const revertedTransactionCumulativeGasUsed =
         cumulativeGasUsages.find((gas, i) => {
           return (
-            Number(transactionWithReceiptExtended.ethTx.transactionIndex) >=
+            Number(transactionWithReceiptExtended.ethTx!.transactionIndex) >=
               cumulativeGasUsages.length - 1 - i && gas
           );
         }) ?? 0n;
 
       const ethReceipt = toRevertedOutOfResourcesReceipt({
-        transaction: transactionWithReceiptExtended.ethTx,
+        transaction: transactionWithReceiptExtended.ethTx!,
         blockNumber,
         blockHash,
         cumulativeGasUsed: revertedTransactionCumulativeGasUsed,
@@ -270,7 +274,7 @@ export default async function transform({
   filteredTransactions.forEach((transaction) => {
     store.push({
       collection: Collection.Transactions,
-      data: { tx: transaction.ethTx },
+      data: { tx: transaction.ethTx! },
     });
     store.push({
       collection: Collection.Receipts,
--- a/indexer/src/main.ts
+++ b/indexer/src/main.ts
@@ -83,7 +83,7 @@ export default async function transform({
   const store: Array<StoreItem> = [];
   const blockLogsBloom = new Bloom();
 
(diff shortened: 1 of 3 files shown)


indexer/src/main.ts
```diff
@@ -82,112 +82,126 @@ export default async function transform({
...
+    // Number(null) is 0 so this won't panic.
+    cumulativeGasUsages[Number(eventExtended.ethTx.transactionIndex)] =
+      cumulativeGasUsed;
 
-      // Add the logs bloom of the receipt to the block logs bloom.
-      const receiptLogsBloom = new Bloom(hexToBytes(ethReceipt.logsBloom));
-      blockLogsBloom.or(receiptLogsBloom);
+    // Add all the eth data to the store.
+    store.push({
+      collection: "transactions",
```
tcoratger, line 178: Can we have collection name constants in our constants file and simply use these constants in order to avoid typo or confusions?

indexer/src/main.ts
```diff
@@ -82,112 +82,126 @@ export default async function transform({
...
+      data: { tx: eventExtended.ethTx },
+    });
+    store.push({
+      collection: "receipts",
```
tcoratger, line 182: Same remark here and for all places in the indexer where we have hardcoded collection names

indexer/src/main.ts
```diff
@@ -208,55 +222,59 @@ export default async function transform({
...
       // const cumulativeGasUsages = [300n, undefined, undefined, 200n, undefined, 100n, undefined, undefined, 10n, undefined];
       // const ethTx = { transactionIndex: 5 };
       // const revertedTransactionCumulativeGasUsed = 100n;
-      const len = cumulativeGasUsages.length;
       const revertedTransactionCumulativeGasUsed =
         cumulativeGasUsages.find((gas, i) => {
           return (
-            Number(ethTx.transactionIndex) >= len - 1 - i && gas !== undefined
+            Number(transactionWithReceiptExtended.ethTx.transactionIndex) >=
+              cumulativeGasUsages.length - 1 - i && gas !== undefined
```
tcoratger, line 252:
```suggestion
              cumulativeGasUsages.length - 1 - i && gas
```

tcoratger:
Hey @cwastche can you have a look at the failing tests?

cwastche:
Fixed typing issues, tests seem to pass when ran locally

indexer/src/main.ts
```diff
@@ -82,112 +82,130 @@ export default async function transform({
...
-      // 1. The transaction is missing calldata.
-      // 2. The transaction is a multi-call.
-      // 3. The length of the signature array is different from 5.
-      // 4. The chain id is not encoded in the v param of the signature for a
-      //    Legacy transaction.
-      // 5. The deserialization of the transaction fails.
-      if (!typedEthTx) {
-        return null;
-      }
+  const filteredEvents = (events ?? [])
```
eugypalu, line 93: can we improve the type safety here?

indexer/src/main.ts
```diff
@@ -82,112 +82,130 @@ export default async function transform({
...
 
-      // Can be null if:
-      // 1. The event is part of the defined ignored events (see IGNORED_KEYS).
-      // 2. The event has an invalid number of keys.
-      const ethLogs = receipt.events
-        .map((e) => {
-          return toEthLog({
-            transaction: ethTx,
+      return { ...eventExtended, ethTx };
+    })
```
eugypalu, line 121: Do you think it might be possible to break the code using helper functions to make it more readable?

indexer/src/main.ts
```diff
@@ -73,206 +79,238 @@ export default async function transform({
...
+
+function accumulateGasAndUpdateStore(
+  processedEvents: ProcessedEvent[],
+  store: Array<StoreItem>,
+  blockLogsBloom: Bloom,
+): { cumulativeGasUsed: bigint; cumulativeGasUsages: bigint[] } {
+  let cumulativeGasUsed = 0n;
+  const cumulativeGasUsages: bigint[] = [];
+
+  if (!Array.isArray(processedEvents)) {
```
ClementWalter, line 196: how can `processedEvents` not be an array since the type is `ProcessedEvent[]` ?

indexer/src/main.ts
```diff
@@ -73,206 +79,238 @@ export default async function transform({
...
+    event.ethReceipt.cumulativeGasUsed = `0x${cumulativeGasUsed.toString(16)}`;
+
+    updateStore(store, event);
+    updateBlockLogsBloom(blockLogsBloom, event);
+  });
+
+  return { cumulativeGasUsed, cumulativeGasUsages };
+}
+
+function updateStore(store: Array<StoreItem>, event: ProcessedEvent) {
```
ClementWalter, line 214: I don't think that this kind of one-time-used mutating functions are necessary

indexer/src/main.ts
```diff
@@ -73,206 +79,238 @@ export default async function transform({
...
+    collection: Collection.Transactions,
+    data: { tx: event.ethTx },
+  });
+  store.push({
+    collection: Collection.Receipts,
+    data: { receipt: event.ethReceipt },
+  });
+  event.ethLogs.forEach((log) =>
+    store.push({ collection: Collection.Logs, data: { log } })
+  );
```
ClementWalter, line 225:
```suggestion
  store.push([
  {
    collection: Collection.Transactions,
    data: { tx: event.ethTx },
  },
  {
    collection: Collection.Receipts,
    data: { receipt: event.ethReceipt },
  },
  ...event.ethLogs.map((log) =>({ collection: Collection.Logs, data: { log } }))
  ])
```

indexer/src/main.ts
```diff
@@ -73,206 +79,238 @@ export default async function transform({
...
+    // 4. The chain id is not encoded in the v param of the signature for a
+    //    Legacy transaction.
+    // 5. The deserialization of the transaction fails.
+    if (!typedEthTx) return null;
+
+    const ethTx = typedTransactionToEthTx({
+      typedTransaction: typedEthTx!,
+      receipt: event.receipt,
+      ...blockInfo,
+    });
```
ClementWalter, line 159: there is a naming mistmatch here. We first create a "eth" tx, then say that it's a "typed" tx (which also looks like a eth tx to me) so that eventually it becomes a "eth" tx again

indexer/src/main.ts
```diff
@@ -73,206 +79,238 @@ export default async function transform({
...
 }
+
+function createBlockInfo(header: BlockHeader): BlockInfo {
+  const blockNumber = padString(toHexString(header.blockNumber), 8);
+  const blockHash = padString(header.blockHash, 32);
+  const isPendingBlock = blockHash === NULL_BLOCK_HASH;
+  return { blockNumber, blockHash, isPendingBlock };
+}
+
+function processEvent(blockInfo: BlockInfo) {
```
ClementWalter, line 143: do we need the two initial early return as we already apply filters before

```
    .filter((event) => isKakarotTransaction(event.transaction))
    // Skip if the transaction_executed event contains "eth validation failed".
    .filter((event) => !ethValidationFailed(event.event))
```

indexer/src/main.ts
```diff
@@ -73,206 +79,238 @@ export default async function transform({
...
+    collection: Collection.Headers,
     data: { header: ethHeader },
   });
 
   return store;
 }
+
+function createBlockInfo(header: BlockHeader): BlockInfo {
+  const blockNumber = padString(toHexString(header.blockNumber), 8);
+  const blockHash = padString(header.blockHash, 32);
```
ClementWalter, line 138: how does this differ from native `.padStart` or `.padEnd` ?

eugypalu: padString adds a "0x" prefix and handles default values, ensuring proper hexadecimal formatting, while native .padStart and .padEnd only pad strings without any additional logic

indexer/src/main.ts
eugypalu: The function toTypedEthTx converts the transaction into a typed Ethereum transaction, returning an object of type TypedTransaction. 
On the other hand, typedTransactionToEthTx takes this TypedTransaction and converts it into the Ethereum transaction format, returning an object of type ExtendedJsonRpcTx.

In both cases, they are eth transactions, but they return objects of different types, which is why those names were chosen. What do you think?

indexer/src/main.ts
eugypalu: fixed, thanks

ClementWalter (approved):

PR #1315, opened 2024-07-28
bump rust to 1.80 and rm lazy_static

tcoratger:
Related:
 - In Rust 1.80, `LazyCell` and `LazyLock` are stable https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#lazycell-and-lazylock, giving us the opportunity to remove an useless dependency (`static_lazy`).
 - Similar issue is raised in reth: https://github.com/paradigmxyz/reth/issues/9799
 - `lazy_static` appears to be intended to be deprecated https://github.com/rust-lang-nursery/lazy-static.rs/issues/214

### DIFF
--- a/src/bin/katana_genesis.rs
+++ b/src/bin/katana_genesis.rs
@@ -1,21 +1,24 @@
 use dotenvy::dotenv;
 use kakarot_rpc::test_utils::katana::genesis::KatanaGenesisBuilder;
-use lazy_static::lazy_static;
 use reth_primitives::{B256, U256};
 use starknet::core::types::Felt;
 use std::{
     env::var,
     path::{Path, PathBuf},
     str::FromStr,
+    sync::LazyLock,
 };
 
-lazy_static! {
-    static ref GENESIS_FOLDER_PATH: PathBuf = Path::new(env!("CARGO_MANIFEST_DIR")).to_path_buf().join(".katana");
-    static ref KAKAROT_CONTRACTS_PATH: PathBuf =
-        Path::new(env!("CARGO_MANIFEST_DIR")).to_path_buf().join("lib/kakarot/build");
-    static ref COINBASE_ADDRESS: Felt = 0x12345u32.into();
-    static ref SALT: Felt = Felt::ZERO;
-}
+/// Katana genesis folder path.
+static GENESIS_FOLDER_PATH: LazyLock<PathBuf> =
+    LazyLock::new(|| Path::new(env!("CARGO_MANIFEST_DIR")).to_path_buf().join(".katana"));
+
+/// Kakarot contracts path.
+static KAKAROT_CONTRACTS_PATH: LazyLock<PathBuf> =
+    LazyLock::new(|| Path::new(env!("CARGO_MANIFEST_DIR")).to_path_buf().join("lib/kakarot/build"));
+
+/// Mock coinbase address.
+static COINBASE_ADDRESS: LazyLock<Felt> = LazyLock::new(|| 0x12345u32.into());
 
 fn main() {
     // Load the env vars.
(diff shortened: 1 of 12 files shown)


src/bin/katana_genesis.rs
```diff
@@ -1,21 +1,24 @@
...
-lazy_static! {
-    static ref GENESIS_FOLDER_PATH: PathBuf = Path::new(env!("CARGO_MANIFEST_DIR")).to_path_buf().join(".katana");
-    static ref KAKAROT_CONTRACTS_PATH: PathBuf =
-        Path::new(env!("CARGO_MANIFEST_DIR")).to_path_buf().join("lib/kakarot/build");
-    static ref COINBASE_ADDRESS: Felt = 0x12345u32.into();
-    static ref SALT: Felt = Felt::ZERO;
-}
+/// Katana genesis folder path.
+static GENESIS_FOLDER_PATH: LazyLock<PathBuf> =
+    LazyLock::new(|| Path::new(env!("CARGO_MANIFEST_DIR")).to_path_buf().join(".katana"));
```
Eikix, line 14: What is a LazyLock? Is it the new lazy_static syntax?

Eikix:
One question to approve this PR, are there any repos in the ecosystem that have bumped to rust 1.80? Will it break something?

Have these repos also replaced lazy_static by LazyLock?

src/bin/katana_genesis.rs
tcoratger: This is the new stable feature of Rust 1.80: https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#lazycell-and-lazylock

tcoratger:
> One question to approve this PR, are there any repos in the ecosystem that have bumped to rust 1.80? Will it break something?
> 
> Have these repos also replaced lazy_static by LazyLock?

Yes reth here: https://github.com/paradigmxyz/reth/issues/9799

tcoratger:
> ?

They had `once_cell::sync::Lazy` which is equivalent but not a macro. Also I read this issue https://github.com/rust-lang-nursery/lazy-static.rs/issues/214 which seems to favor a deprecation of `lazy_static` and now that the equivalent feature is stable on std, I fear that this deprecation will accelerate so I wanted to move forward before. 

But it is continuous improvement also done to remove an additional dependency. If ever we can wait for Greg to have a third opinion on the question :)

Eikix:
> > ?
> 
> They had `once_cell::sync::Lazy` which is equivalent but not a macro. Also I read this issue [rust-lang-nursery/lazy-static.rs#214](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214) which seems to favor a deprecation of `lazy_static` and now that the equivalent feature is stable on std, I fear that this deprecation will accelerate so I wanted to move forward before.
> 
> But it is continuous improvement also done to remove an additional dependency. If ever we can wait for Greg to have a third opinion on the question :)

Last question, we've had issues with bumping rust version in the past, is this fine wrt our dependencies?

tcoratger:
> > > ?
> > 
> > 
> > They had `once_cell::sync::Lazy` which is equivalent but not a macro. Also I read this issue [rust-lang-nursery/lazy-static.rs#214](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214) which seems to favor a deprecation of `lazy_static` and now that the equivalent feature is stable on std, I fear that this deprecation will accelerate so I wanted to move forward before.
> > But it is continuous improvement also done to remove an additional dependency. If ever we can wait for Greg to have a third opinion on the question :)
> 
> Last question, we've had issues with bumping rust version in the past, is this fine wrt our dependencies?

Should be fine and as reth will bump, as always if we want to continue using reth we have to bump too... but we could put this pr on hold until we need to bump reth for ex to be 100% sure we absolutely need it.

Eikix (approved):
lgtm

Eikix:
Continuous improvement doesn't wait. 👑

Issue #1349, opened 2024-08-28
feat: add unit tests for the mempool in `send_raw_transaction`


tcoratger:
### Test `send_raw_transaction` with Different Scenarios

**Context:**

Now that https://github.com/kkrt-labs/kakarot-rpc/pull/1321 is merged, we need to thoroughly test the `send_raw_transaction` method to ensure that the validation inside the mempool is functioning correctly.

**Objective:**

We aim to verify that the mempool:

1. Correctly fills up with transactions that meet the required conditions (e.g., sufficient balance, correct signature, correct chain ID, etc.).
2. Properly rejects invalid transactions across various scenarios, covering all possible inconsistencies.

**Tasks:**

- [ ] Write unit tests to validate that transactions with correct conditions are successfully added to the mempool.
- [ ] Write unit tests to ensure that transactions failing to meet the criteria (e.g., insufficient balance, incorrect signature, wrong chain ID) are correctly rejected by the mempool.
- [ ] Validate the overall integration of the `send_raw_transaction` method with the mempool, ensuring that both acceptance and rejection flows are handled as expected.

**Outcome:**

These unit tests will confirm the correct implementation of the mempool's validation logic and its integration within the `send_raw_transaction` method.

PR #1354, opened 2024-08-29
feat: add unit tests for the mempool in send_raw_transaction

eugypalu:
Resolves: #1349 

# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Testing

# What is the new behavior?

unit tests for the send_raw_transaction method, ensuring that the mempool correctly validates and handles transactions, including both acceptance of valid transactions and rejection of invalid ones

- [ ] Yes
- [x] No

### DIFF
--- a/tests/tests/eth_provider.rs
+++ b/tests/tests/eth_provider.rs
@@ -839,7 +839,7 @@ async fn test_send_raw_transaction_wrong_signature(#[future] katana: Katana, _se
 #[rstest]
 #[awt]
 #[tokio::test(flavor = "multi_thread")]
-async fn test_send_transaction_wrong_chain_id(#[future] katana: Katana, _setup: ()) {
+async fn test_send_raw_transaction_wrong_chain_id(#[future] katana: Katana, _setup: ()) {
     // Given
     let eth_provider = katana.eth_provider();
     let wrong_chain_id = 999; // An arbitrary wrong chain ID
@@ -877,7 +877,7 @@ async fn test_send_transaction_wrong_chain_id(#[future] katana: Katana, _setup: 
 #[rstest]
 #[awt]
 #[tokio::test(flavor = "multi_thread")]
-async fn test_send_transaction_insufficient_balance(#[future] katana: Katana, _setup: ()) {
+async fn test_send_raw_transaction_insufficient_balance(#[future] katana: Katana, _setup: ()) {
     // Given
     let eth_provider = katana.eth_provider();
     let eoa = katana.eoa();
@@ -889,7 +889,7 @@ async fn test_send_transaction_insufficient_balance(#[future] katana: Katana, _s
         nonce: 0,
         gas_limit: 21000,
         to: TxKind::Call(Address::random()),
-        value: U256::from(1_000_000_001),
+        value: U256::MAX,
         max_fee_per_gas: 875_000_000,
         max_priority_fee_per_gas: 0,
         input: Bytes::default(),
@@ -904,10 +904,7 @@ async fn test_send_transaction_insufficient_balance(#[future] katana: Katana, _s
     let mempool_size = eth_provider.mempool().unwrap().pool_size();
 
     // Attempt to send the transaction
-    let result = eth_provider.send_raw_transaction(transaction_signed.envelope_encoded()).await;
-
-    // Then
-    assert!(result.is_err()); // Ensure the transaction is rejected
+    let _ = eth_provider.send_raw_transaction(transaction_signed.envelope_encoded()).await;
 
     // Verify that the number of pending transactions in the mempool remains unchanged
     assert_eq!(eth_provider.mempool().unwrap().pool_size().pending, mempool_size.pending);
--- a/tests/tests/eth_provider.rs
+++ b/tests/tests/eth_provider.rs
@@ -889,7 +889,7 @@ async fn test_send_transaction_insufficient_balance(#[future] katana: Katana, _s
         nonce: 0,
         gas_limit: 21000,
         to: TxKind::Call(Address::random()),
-        value: U256::MAX, // Exceeding balance
+        value: U256::from(1_000_000_001),
         max_fee_per_gas: 875_000_000,
         max_priority_fee_per_gas: 0,
         input: Bytes::default(),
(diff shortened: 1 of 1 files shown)


tests/tests/eth_provider.rs
```diff
@@ -725,6 +726,9 @@ async fn test_send_raw_transaction(#[future] katana: Katana, _setup: ()) {
     let signature = sign_message(katana.eoa().private_key(), transaction.signature_hash()).unwrap();
     let transaction_signed = TransactionSigned::from_transaction_and_signature(transaction, signature);
 
+    // Retrieve the current size of the mempool
+    let mempool_size = eth_provider.mempool().unwrap().pool_size();
+
```
tcoratger, line 731: Can we check that the size of the mempool is indeed 0 at the beginning here?

tests/tests/eth_provider.rs
```diff
@@ -743,6 +747,11 @@ async fn test_send_raw_transaction(#[future] katana: Katana, _setup: ()) {
     // Assert the transaction hash and block number
     assert_eq!(tx.hash, transaction_signed.hash());
     assert!(tx.block_number.is_none());
+
+    // Assert that the number of pending transactions in the mempool has increased by 1
+    assert_eq!(eth_provider.mempool().unwrap().pool_size().pending, mempool_size.pending + 1);
+    // Assert that the transaction in the mempool exists
+    assert!(eth_provider.mempool().unwrap().get(&tx.hash).is_some());
```
tcoratger, line 754: Nice test, can we check additionally here that the recovered hash is `tx.hash` (the hash that we want)?

tests/tests/eth_provider.rs
```diff
@@ -810,6 +819,9 @@ async fn test_send_raw_transaction_wrong_signature(#[future] katana: Katana, _se
     // Set an incorrect signature
     transaction_signed.signature = Signature::default();
 
+    // Retrieve the current size of the mempool
+    let mempool_size = eth_provider.mempool().unwrap().pool_size();
+
```
tcoratger, line 824: Same can we also assert the zero size of the mempool here?

tests/tests/eth_provider.rs
```diff
@@ -819,6 +831,83 @@ async fn test_send_raw_transaction_wrong_signature(#[future] katana: Katana, _se
 
     // Assert that no transaction is found
     assert!(tx.is_none());
+
+    // Verify that the number of pending transactions in the mempool remains unchanged
+    assert_eq!(eth_provider.mempool().unwrap().pool_size().pending, mempool_size.pending);
```
tcoratger, line 836: Can we also check the transaction hash that is inside the mempool here?

tests/tests/eth_provider.rs
```diff
@@ -819,6 +831,83 @@ async fn test_send_raw_transaction_wrong_signature(#[future] katana: Katana, _se
...
+    let mempool_size = eth_provider.mempool().unwrap().pool_size();
+
+    // Attempt to send the transaction
+    let result = eth_provider.send_raw_transaction(transaction_signed.envelope_encoded()).await;
+
+    // Then
+    assert!(result.is_err()); // Ensure the transaction is rejected
+
+    // Verify that the number of pending transactions in the mempool remains unchanged
+    assert_eq!(eth_provider.mempool().unwrap().pool_size().pending, mempool_size.pending);
```
tcoratger, line 874: I think that here we should assert that the size is zero to really be sure that there is nothing to pollute the pool inside (we called `send_raw_transaction` only once and the conditions were invalid so there should be nothing in the mempool at all)

tests/tests/eth_provider.rs
```diff
@@ -819,6 +831,83 @@ async fn test_send_raw_transaction_wrong_signature(#[future] katana: Katana, _se
...
+    let transaction_signed = TransactionSigned::from_transaction_and_signature(transaction, signature);
+
+    // Retrieve the current size of the mempool
+    let mempool_size = eth_provider.mempool().unwrap().pool_size();
+
+    // Attempt to send the transaction
+    let _ = eth_provider.send_raw_transaction(transaction_signed.envelope_encoded()).await;
+
+    // Verify that the number of pending transactions in the mempool remains unchanged
+    assert_eq!(eth_provider.mempool().unwrap().pool_size().pending, mempool_size.pending);
```
tcoratger, line 910: Same here can we assert 0 size?

tcoratger:
Can we also test here:
- The `transaction_size > self.max_tx_input_bytes` condition by providing an artificially huge number of bytes for the transaction request to verify that it does not enter the pool (preventing a DOS attack)
- The `transaction_gas_limit > self.block_gas_limit` condition with a transaction where the gas limit is artificially huge to verify that the transaction does not enter the pool
- `transaction.max_priority_fee_per_gas() > Some(transaction.max_fee_per_gas())` condition
- The `transaction.nonce() < account.nonce` condition by providing an artificial wrong nonce to the transaction?

tests/tests/eth_provider.rs
tcoratger: Each time also can we check the `total` of the pool size https://github.com/paradigmxyz/reth/blob/019ec727a2e078fd1517d5b04f39f6713d3bf0d7/crates/transaction-pool/src/traits.rs#L1259-L1280 to be sure that everything has been set up correctly in the pending pool and that nothing has been entered by mistake, for example in the blob pool or in the base fee pool

tests/tests/eth_provider.rs
eugypalu: agree, fixed here and in all the other tests

tests/tests/eth_provider.rs
eugypalu: fixed, thanks

tests/tests/eth_provider.rs
eugypalu: Agree, I’ve added the check on the total size of the mempool, while I didn’t add the hash check here because the mempool is empty, as the signature is incorrect. What do you think?

eugypalu:
> Can we also test here:
> 
> * The `transaction_size > self.max_tx_input_bytes` condition by providing an artificially huge number of bytes for the transaction request to verify that it does not enter the pool (preventing a DOS attack)
> * The `transaction_gas_limit > self.block_gas_limit` condition with a transaction where the gas limit is artificially huge to verify that the transaction does not enter the pool
> * `transaction.max_priority_fee_per_gas() > Some(transaction.max_fee_per_gas())` condition
> * The `transaction.nonce() < account.nonce` condition by providing an artificial wrong nonce to the transaction?

Added all the tests, thanks

tcoratger (approved):

PR #1376, opened 2024-09-11
Add trunk to CI and fixes

ClementWalter:


.github/workflows/rust_test.yml
```diff
@@ -5,17 +5,6 @@ on:
...
-    name: fmt
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
-    steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@nightly
-        with:
-          components: rustfmt
-      - run: cargo fmt --all --check
-
```
tcoratger, line 18: Are we sure that the trunk fmt check is done properly with cargo nightly? because this is important to sort the imports properly in each file, which is not done in non-nightly mode

ClementWalter: copied from last keth update, trunk uses the system installed (meaning the project) rust

ClementWalter: confirming that this
image
is formatted into 
image
by `trunk fmt`

tcoratger: nice

greged93 (approved):

PR #1452, opened 2024-10-11
bump sink-mongo to 0.9.2

tcoratger:
Should close https://github.com/kkrt-labs/kakarot-rpc/issues/1433


tcoratger:
> @tcoratger you need to compute the new nix path for the updated sink-mongo image

I always forget to do that even if this is written as a comment ahah, done

greged93 (approved):

PR #1517, opened 2024-10-30
set block gas limit in pool to 7M

tcoratger:
Should close https://github.com/kkrt-labs/kakarot-rpc/issues/1479


Eikix (commented):
can we test this?

Makefile
```diff
@@ -62,6 +62,9 @@ test-ci: load-env
 test-target: load-env
 	cargo test --tests --all-features $(TARGET) -- --nocapture
 
+test-target1: load-env
```
Eikix, line 65: why do we add a new target here?

Makefile
```diff
@@ -62,6 +62,9 @@ test-ci: load-env
...
+	cargo test --package kakarot-rpc --test entry --all-features -- tests::mempool::test_maintain_mempool --exact --show-output
+
```
tcoratger, line 67:
```suggestion
```

Makefile
tcoratger: not needed, for tests, removed

greged93 (approved):

PR #1530, opened 2024-11-06
invoke V3

### LINKED ISSUE #773: feat: upgrade Starknet tx type to v3 and pay fees in STRK
# Feature Request

**Describe the Feature Request**

To simplify our testnet fee strategy, we'll use the v3 tx type and pay fees in STRK.
At the SW appchain level, the STRK fees will be set to 0.

```[tasklist]
### Tasks
- [ ] make sure starknet rs is bumped to 0.9
- [ ] make sure katana can handle v3 txs and that the --disable-fee flag of katana sets both STRK and ETH fees to 0
- [ ] change transaction type in add raw transaction route to be v3 and pay in STRK
- [ ] set MAX_FEE in the tx v3 field to 0
```

jbcaron:
Time spent on this PR:

Resolves: #773 

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

# What is the new behavior?

- [ ] Yes
- [ ] No

### DIFF
--- a/src/models/felt.rs
+++ b/src/models/felt.rs
@@ -52,6 +52,28 @@ impl TryFrom<Felt252Wrapper> for Address {
     }
 }
 
+impl TryFrom<Felt252Wrapper> for u64 {
+    type Error = EthereumDataFormatError;
+
+    fn try_from(felt: Felt252Wrapper) -> Result<Self, Self::Error> {
+        match felt.to_be_digits() {
+            [0, 0, 0, d] => Ok(d),
+            _ => Err(EthereumDataFormatError::Primitive),
+        }
+    }
+}
+
+impl TryFrom<Felt252Wrapper> for u128 {
+    type Error = EthereumDataFormatError;
+
+    fn try_from(felt: Felt252Wrapper) -> Result<Self, Self::Error> {
+        match felt.to_be_digits() {
+            [0, 0, d1, d2] => Ok(Self::from(d1) << 64 | Self::from(d2)),
+            _ => Err(EthereumDataFormatError::Primitive),
+        }
+    }
+}
+
 impl From<B256> for Felt252Wrapper {
     fn from(value: B256) -> Self {
         Self(Felt::from_bytes_be(value.as_ref()))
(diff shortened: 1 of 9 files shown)


Eikix:
It seems for rust tests, the katana accounts are not funded in STRK?

src/providers/eth_provider/starknet/mod.rs
```diff
@@ -4,10 +4,13 @@ pub mod relayer;
...
 abigen_legacy!(ERC20, "./.kakarot/artifacts/ERC20.json");
 
 /// Starknet native token address
-pub static STARKNET_NATIVE_TOKEN: LazyLock<Felt> =
-    LazyLock::new(|| Felt::from_hex("0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7").unwrap());
+pub static STARKNET_NATIVE_TOKEN: Felt =
+    Felt::from_hex_unchecked("0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d");
+
+/// Starknet parent token address
+pub static STARKNET_PARENT_TOKEN: Felt =
```
Eikix, line 15: Can we name this STARKNET_ETH

src/providers/eth_provider/state.rs
```diff
@@ -66,7 +66,10 @@ where
         // Convert the optional Ethereum block ID to a Starknet block ID.
         let starknet_block_id = self.to_starknet_block_id(block_id).await?;
         // Get the balance of the address at the given block ID.
-        self.starknet_provider().balance_at(starknet_address(address), starknet_block_id).await.map_err(Into::into)
+        self.starknet_provider()
+            .balance_at_native(starknet_address(address), starknet_block_id)
```
Eikix, line 70: I think you're replacing it wrongly
This should still be ETH here.

There is a difference between relayers (workers that relay EVM transactions) -> they pay in STRK and their balances are STRK

versus:

Kakarot users -> their native token remains ETH

src/providers/sn_provider/starknet_provider.rs
```diff
@@ -38,25 +38,55 @@ where
...
+    /// * `block_id` - The block ID at which to check the balance
     ///
-    /// If the contract is not deployed or the class hash is not declared, a balance of 0 is returned
-    /// instead of an error.
-    pub async fn balance_at(&self, address: Felt, block_id: BlockId) -> Result<U256, ExecutionError> {
-        // Create a new `ERC20Reader` instance for the Starknet native token
-        let eth_contract = ERC20Reader::new(*STARKNET_NATIVE_TOKEN, &self.provider);
+    /// # Returns
+    /// * `Result<U256, ExecutionError>` - The balance as a U256 value or an execution error
+    pub async fn balance_at_parent(&self, address: Felt, block_id: BlockId) -> Result<U256, ExecutionError> {
```
Eikix, line 50: i don't think we should call ETH parent in terms of naming

We could make in unambiguous by just calling it: eth_balance_at and strk_balance_at

Eikix: Ok wait nevermind 
you are implying that since this codebase can be used by any starknet appchain
some appchains might want to use their own token as base token

Eikix: at any rate, since STARKNET_PARENT_TOKEN is a fixed address and not even an injected variable. For now it only ever points to ETH
So I think my point remains valid

If one day we want to parametrize base token, we will do it

src/providers/eth_provider/state.rs
jbcaron: Got it, I understand.

src/models/felt.rs
```diff
@@ -52,6 +52,28 @@ impl TryFrom<Felt252Wrapper> for Address {
...
+impl TryFrom<Felt252Wrapper> for u128 {
+    type Error = EthereumDataFormatError;
+
+    fn try_from(felt: Felt252Wrapper) -> Result<Self, Self::Error> {
+        match felt.to_be_digits() {
+            [0, 0, d1, d2] => Ok(Self::from(d1) << 64 | Self::from(d2)),
+            _ => Err(EthereumDataFormatError::Primitive),
+        }
+    }
+}
```
tcoratger, line 75: Can't we use directly this

https://github.com/starknet-io/types-rs/blob/7ef30234da7957f7585ea832d57d13083c76995c/crates/starknet-types-core/src/felt/primitive_conversions.rs#L91-L96

by doing something like

```rust
let a = Felt252Wrapper;
let b: u64 = a.0.try_into().expect("");
```

PR #1541, opened 2024-11-19
Update Kakarot to v0.9.8

ClementWalter:
Update Kakarot to v0.9.8

### DIFF
--- a/.github/workflows/rust_test.yml
+++ b/.github/workflows/rust_test.yml
@@ -25,7 +25,7 @@ jobs:
       - name: install foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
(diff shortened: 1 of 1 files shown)


enitrat:
I think it might be an issue on staging to not have  https://github.com/kkrt-labs/kakarot/issues/1629

.github/workflows/rust_test.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: install foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
+          version: nightly-a79dfaed6fc6f88cda5f314a25d1b484d9d8c051
```
obatirou, line 28: This should not be necessary anymore 
The issue is closed and should be fixed normally https://github.com/foundry-rs/foundry/issues/9349
cc @enitrat

